Issue #15, opened 2015-06-10
Execution error of some examples when tracing of msg processes is enabled


mpoquet:
Hello,
I'm currently creating a MSG-based simulator. When I enable tracing (with --cfg=tracing:1 --cfg=tracing/msg/process:1) I get an exception saying that a container (corresponding to a process) cannot be found.

As a MWE, I found that the example masterslave_kill.c (examples/msg/masterslave) crashes in the same way. Some steps to reproduce the error are given above.

cd SG_DEV_ROOT
cd examples/msg/masterslave
gcc -g -lsimgrid -o msk masterslave_kill.c
./msk ../../platforms/small_platform.xml deployment_masterslave_kill.xml --cfg=tracing:1 --cfg=tracing/msg/process:1

mpoquet:
The problem seems to be in file msg_process.c at line 174. The crash seems to occur when, in the simcall_process_create call (just above), SIMIX decides to execute the process instantly.

schnorr:
I was able to reproduce the problem here:

$ ./msk  ../../platforms/small_platform.xml deployment_masterslave_kill.xml --cfg=tracing:1 --cfg=tracing/msg/process:1
[0.000000] [xbt_cfg/INFO] Configuration change: Set 'tracing' to '1'
[0.000000] [xbt_cfg/INFO] Configuration change: Set 'tracing/msg/process' to '1'
[Tremblay:master:(1) 0.000000] [msg_test/INFO] Hello!
[Fafard:slave:(2) 0.000000] [msg_test/INFO] Hello!
[Fafard:slave:(2) 0.000000] [msg_test/INFO] Suspending myself
** SimGrid: UNCAUGHT EXCEPTION received on ./msk(2): category: tracing error; value: 1
** container with name slave-2 not found
** Thrown by slave() in this process
[Fafard:slave:(2) 0.000000] /home/schnorr/svn/simgrid/src/xbt/ex.c:148: [xbt_ex/CRITICAL] container with name slave-2 not found

**   In PJ_container_get() at /home/schnorr/svn/simgrid/src/instr/instr_paje_containers.c:145
**   In TRACE_msg_process_suspend.part.5() at /home/schnorr/svn/simgrid/src/msg/instr_msg_process.c:91
**   In MSG_process_suspend() at /home/schnorr/svn/simgrid/src/msg/msg_process.c:440
**   In slave() at /home/schnorr/svn/simgrid/examples/msg/masterslave/masterslave_kill.c:21
[...]

PR #17, opened 2015-06-10
Candidate patch to issue #15

mpoquet:


mquinson:
@schnorr could you please have a look at that pull request? It is ways above my current understanding of the tracing subsystem..

Thanks, Mt

mquinson:
@alegrand  is a jerk

mquinson:
@schnorr could you please review that change, please?

schnorr:
This commit solves a problem I was expecting to appear at some time in the future, regarding the decision of immediate execution of a Simix process. We were tracing its creation only after the Simix call, but that is dangerous since Simix might decide to run that process to the end immediately. So when the respective process container is going to be destroyed, it notices that it was never created. It looks like someone stumbled upon this problem and the proposed fix definitely solves the issue (I was able to reproduce the problem and the fix).

The solution was to create the process trace container before the Simix call. There might be still a problem here if eventually Simix decides to not create the process. If that eventually happens, we'll have a process creation that shouldn't be and the respective destroy will never come. I am not sure how often Simix decides to not create a process, but I believe the trace will survive on this scenario.

My final comment on this is that the Trace requires the pid of the process to create it. The solution proposed by @mpoquet is to use the function SIMIX_process_get_maxpid() to get the PID that will be used by  simcall_process_create to create the process. If that behavior changes in the future, it will definitely break the tracing of process creation when tracing/msg/process is enabled. But then we'll fix.

For such reasons, I recommend merging this commit. If you do, you'll have to include put this line to avoid a warning being treated as error in Trace CI:
#include "simix/smx_private.h"

@mquinson Review accomplished (on a 14th of July).

mquinson:
Merged. On 14 of July ;)

Thanks for your time,
Mt

mquinson:
As noted in #15, I had to revert this patch as it broke some tests. Please rework your patch and resubmit it!
I would have loved to fix the test myself, but I'm so short on time that I cannot. Sorry

PR #30, opened 2015-10-09
Fix duplicate PTP messages when tracing/smpi/internals is enabled

fernandodeperto:
When the tracing/smpi/internals option is not enabled, some PTP messages
are showed for send and receive operations, both on tracing and
replaying. When the option is enabled, additional messages are
generated, so duplicate messages for send and receive operations
appear on the log.

Initially I thought that the idea was to only show PTP messages when the
tracing/smpi/internals option is enabled. After talking to Arnaud it
seems that is not the case. So I changed the code so both funcionalities
are present: PTP messages for send and receive operations when the
option is disabled and all PTP messages when the option is enabled, as
expected.

I have little knowledge about the internals of the Simgrid code, this fix was
what worked for me in my experiments. It is possible that this is not the ideal
way to fix this.


mquinson:
Thank for the changes. It's merged now.

PR #45, opened 2016-01-07
Patch candidate for issue 15

mpoquet:
This is a patch candidate for issue 15 (discussed [here](https://github.com/mquinson/simgrid/issues/15)).

### DIFF
--- a/src/msg/instr_msg_process.cpp
+++ b/src/msg/instr_msg_process.cpp
@@ -134,10 +134,3 @@ void TRACE_msg_process_sleep_out(msg_process_t process)
     new_pajePopState (MSG_get_clock(), process_container, type);
   }
 }
-
-void TRACE_msg_process_end(msg_process_t process)
-{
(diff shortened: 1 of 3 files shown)


mquinson:
Sorry, but this does not compile, according to travis. Ignore the appveyor issues, windows is broken atm, but please fix the compilation errors. 

I guess that you did not activate the 'enable_compile_warnings' option in ccmake. It gives paranoid compilation flags that we strive using ourselves but don't set by default to decrease the chances of users having problem with those options.

mquinson:
Example of compilation error: https://travis-ci.org/mquinson/simgrid/jobs/100912483

mpoquet:
My bad, I disabled the option because it does not compiler under clang 3.7.0 because override methods are not marked as ones :(.

I switched to gcc 5.3.0 and it seems to compile with enable_compile_warnings after mpoquet/simgrid@4b5d823.

PR #51, opened 2016-01-19
Removes the pid hack introduced in the patch proposal of issue 15

mpoquet:
The pid hack generated errors on some codes so I made a cleaner patch.

The TRACE_msg_process_create is now done in SIMIX_process_create instead of MSG_process_create_with_environment.


mquinson:
Many thanks for this patch. That's indeed much cleaner this way

PR #57, opened 2016-01-31
Adding concurrency to SURF LMM, with oldstyle element set

fabienchaix:
Modification to the SURF LMM:
-> Enabling concurrency control and monitoring 
-> Improving efficiency of lmm_update_modified_set
-> Removing  dead code and adding comments

### DIFF
--- a/src/include/surf/maxmin.h
+++ b/src/include/surf/maxmin.h
@@ -49,6 +49,7 @@
  * \f$\sum_if(x_i)\f$, where \f$f\f$ is a strictly increasing concave
  * function.
  *
+ *
  * Constraint: 
  *  - bound (set)
  *  - shared (set)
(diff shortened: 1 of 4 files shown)


mquinson:
Would you say that we can ignore the failure on the smpi-replay-multiple test? In other words, is there any justification in your change for the small timing changes that can be observed? We may have discussed this, but I forgot, sorry. And it's probably better to discuss it on this public bug tracker.

fabienchaix:
Yes, the small difference regarding termination time in this test is due to the serialization happening in this test, because of the introduction of the concurrency limit. 
IMO, this is actually good that at least one test do show a small variation in termination time.
Sorry, I though I had fixed it.

mquinson:
Ah ok. So I merged that. I'll fix the test, no problem. I think we should add some unit testing to this module at some point. I'm currently rewriting the documentation explaining how to do so, so we'll see once I'm done.

mquinson:
I forgot: many thanks for your work, I'm really happy to receive your help with that module that scares me so much!..

mquinson:
It's weird to see that the simulated application goes faster with the serialization: 
  It was 1124646.821338 seconds
It's now 1124382.208410 seconds. 
But it does not really matter: the models were out of charts before your change anyway.

A much more disturbing element is that the simulation time of the smpi-replay-multiple seem to increase quite a bit. I would have to do some more testing to be sure, though. I shouldn't bench code before the first coffee.

fabienchaix:
You are welcome ;-)

I am not sure that I understand what is now faster and what is now slower?

If the actual time it takes to simulate is longer, it means that the code used to stage and unstage variables is taking more time than the savings we get by easing the life of the solving functions, which I suspect is more than linear to the number of enabled variables. 

If the time that is simulated is getting longer, it means that the serialization indeed add some delays. If it getting shorter, it is not unreasonable that we face a "classical anomaly" where pushing too much a resource made it congested, and increased the simulation time. 

Regarding unit testing, I have a prototype essentially using programs already in the testsuite, for the concurrency of network resources, but it would need to be able to configure concurrency from e.g. LUA. If you want to be safe, it would be nice to also have a test from e.g. Julien, for processing resources.

Finally, I have a further commit (spliting element_set and improving a bit the staging mechanism) that I can pull_request when you want.

mquinson:
Ok, I'm done rewriting the documentation about testing. It's here: http://simgrid.gforge.inria.fr/simgrid/3.13/doc/inside_tests.html

That would be really cool if you could add some unit testing to LMM. Please tell me any issue in this documentation, even tiny ones.

As for the other commit, just do the pull request as soon as you are ready. Rotting branches are never good ideas.

PR #59, opened 2016-02-02
Splitting the LMM constraint element_set into enabled_element_set and disabled_element_set

fabienchaix:
In LMM, one of the tricks was that each constraint  used one swag for two sets of elements.
Elements that were enabled (and took part to computations were inserted in the head.
Elements that were disabled (and were discarded) were inserted in the end.
It was not improving readability, and did not save much memory space (2 pointers per element).
It was also making concurrency checks less efficient (because we had to find the first disabled  element by iterating first on the -many- active elements.
It does make the get_var_from_cnst calls more complicated though.


fabienchaix:
@mquinson, in order to test this properly (and have more fun), I would need to get a way to configure concurrency from e.g. LUA...

alegrand:
@fabienchaix you should discuss with @shurakai about the lua integration. He's the one who masters this process.

mquinson:
Well, actually not. Unit tests are located close to the code to test,
both conceptually and in the same file. You want to test only LMM in
the unit tests of LMM, in complete isolation. 

So you cannot use Lua, or your tests will fail when lua is broken. And
if both lua and LMM get mad at the same time, debugging the mess will
be a nightmare.

It would be really satisfying if we could have some tests of LMM in
isolation. Simple tests, non-trivial ones and torture tests, we would
need them all (provided that they are very fast to execute).

That being said, we'll naturally add the lua interface to the new
parts of LMM, for the users and to permit some integration tests (ie,
exercising all parts together).

Thanks for your work,
Mt

-- 
I'm sick of being the guy who eats insects and gets the funny syphilis.

fabienchaix:
@mquinson, regarding the testing of this functionality, you got a point. I propose to modify maxmin_bench, which is basically putting a lot of random constraints and variables together and run the solver once, to also add random concurrency limit to constraints, and different concurrency share to variables, OK?

Also, one of the checks failed with:

 "The command "if [ ${TRAVIS_OS_NAME} = 'linux' ] ; then sudo apt-get install -y cmake valgrind default-jdk gfortran libboost-chrono1.48-dev ; fi" failed and exited with 100 during . "

..which appears as a glitch from Travis to me, Could you trigger a new build?

mquinson:
that's clearly a travis glitch. I just restarted the build.

PR #65, opened 2016-02-25
Adding test for concurrency and cleaner concurrency support

fabienchaix:
This pull request contains a modification of the test of SURF to integrate the tests of concurrency.
It also refactors the way that concurrency is computed, using the generic function lmm_element_concurrency(), which may allow one to define alternative computations, for people who want to create new models.

The test is just a modification of the existing maxmin_bench, while also adding random concurrency_limit and concurrency_shares. I also added a more verbose mode to check the internal state of the maxmin, and a few other  features/corner cases that were not previously stressed in this test.

BTW, I propose to make lmm_print XBT_PUBLIC (instead of XBT_PRIVATE as it is now), so we can actually print the maxmin system (since we give possibility to run it directly). 

Regarding refactoring, I believe it will be very useful for one who wants to implement different models of serialization/limits/low-latency networks, and completely useless for others ;-)
In effect, short discussion shows that not everyone sees the same thing behind the notion of concurrency. So I set the current behaviour to the proposition of Arnaud: Cross-traffic is ignored by concurrency computation, and strange tasks with a "ressource weight" (elem->value) bigger than one still count for one regarding concurrency.


mquinson:
It seems very interesting (even if I didn't carefully checked the diff yet), but you have two broken tests. Could you please comment a bit on this?

Thanks for your work, 
Mt

fabienchaix:
I tried to correct the tests, but I get an assertion failed in Travis, while it works nicely on my computer..
If we want to be fast, we can just forget about this test (the large, non-debug, instances run nicely), but it may worth to investigate a bit?

This is a bit sad that we cannot use debug output because absolute path are inserted by default..
That would definitively help to get coverage up (though it would also mean test more debug code, that we are not so interested in).
One -potentially dirty- way to do that would be to process _FILE_ in the print function to strip the simgrid path (Cmake might even do that). I think the best way would be to compute at compile time how much characters should be stripped, and then the print function just needs to advance the pointer before use.

mquinson:
Unfortunately, we are in a period of frenetic refactoring. I cannot live with a broken test right now. 

I will add an option to strip the code location in debug messages for such situations. Actually, if you specify the log format (as we do in most tests), the same format is also used for debug messages, so they won't show the code location.

In any case, you should add some temporary commits to your pull request, adding all the output you need to understand why mac machines are problematic. That's OK if you break even more tests in these commits. We will just not merge the pull request during the process.

fabienchaix:
Did it ...kind of. I do not use the standard rand(), that seems to behave differently on MAC.
I also had to work to make output robust.

mquinson:
This is great, thanks. Getting the test to be perfectly reproducible is often annoying. In this case, we should maybe use RngStream that we include.

Anyway, you did it very well, thank you.

fabienchaix:
Yep, I expected that there was a way to make portable random numbers in Simgrid, but did not find it in the docs.  
Also, I would like to kindly remind that one will need a way to configure concurrency from LUA (or XML but I understood that we prefer not to add things there)... Would you do it or should I?

mquinson:
I am fighting since about 2 weeks to clean how the resources are created in SimGrid. I want to come up with a clean C++ interface, with no trace of the XML stain. This is not (only) because I am psychorigid, but also because the current interface is so counter-intuitive that it cannot be shown to the users.

I would prefer to wait until after I'm done to add the concurrency to the picture. But you will probably have to remind me again then, as I tend to forget things.

mquinson:
As for RngStream, yeah. That's awfully underdocumented. As usual :(

mquinson:
It seems that we still have some portability issues with the tests that you added. https://ci.inria.fr/simgrid/job/SimGrid-Multi/2745/

I was thinking that sorting the element by names during the output could help, but it will not be sufficient. Maybe you should also do that during the test itself so that the tasks get given to the same resource?

That would be wonderful if you could have a look at that at some point, please. Thanks,

mquinson:
More info: it seems that your code is not portable to 32 bits architectures.

randomstuff:
> Yep, I expected that there was a way to make portable random numbers in Simgrid,
> but did not find it in the docs.

For what's worth, C++11 has standard library for random number generation (`#include <random>`).

fabienchaix:
Yep, actually I copy-pasted one random generator from <random>, since this
source is old-fashion C.

For the portability aspect, indeed I use the generic int, which probably
have a different meaning for 64 and 32 bit...

So, i think the best way out is probably:
1. To use RngStream
2. To check that RngStream uses integers with explicit size
3. Change the random source of RngStream if not
4. Document RngStream :)
Le 29 févr. 2016 11:49 AM, "Gabriel Corona" <notifications@github.com> a
écrit :

mquinson:
If it helps, compiling this code with g++ is very straight forward. I can instruct cmake to do so in a snap.

fabienchaix:
Yep, probably that would help.

But I believe we do not want to use <random> generators from the system, as
they are probably built with the kind of portability we want. Instead I
suggest to use a mere copy of a simple generator in there (and put a big
warning that the random quality is mediocre)
Le 29 févr. 2016 1:08 PM, "Martin Quinson" <notifications@github.com> a
écrit :

mquinson:
one stupid question: the problem seems to be in the test code, but not in the code of maxmin itself, right?

fabienchaix:
Yes. The random sequence is not portable.

However, one of the first bugs was a dirty maxmin fail... Which I cannot
reproduce.
Le 29 févr. 2016 2:26 PM, "Martin Quinson" <notifications@github.com> a
écrit :

PR #66, opened 2016-02-29
Switched maxmin_bench to C++ (in case it helps)

fabienchaix:
Explicitly defined random generator integer type sizes for more portability


randomstuff:
I cherry-picked the commit (minus the C++ switch because I don't think it's necessary) on top of master.

randomstuff:
It might make sense to use unsigned types (`uintXX_t`) instead of the typed ones because overflow (and conversion?) is not undefined behaviour for them.

PR #68, opened 2016-03-16
Energy were not computed correctly for parallel tasks.

glesserd:
Only the energy on the first host of the parallel task was updated.


mquinson:
Stupid me. Thanks for the fix.

PR #69, opened 2016-03-17
Remove useless include that makes projects outside of Simgrid uncompilable.

glesserd:


include/xbt/string.hpp
```diff
@@ -7,7 +7,6 @@
 #ifndef SIMGRID_XBT_STRING_HPP
 #define SIMGRID_XBT_STRING_HPP
 
-#include "src/internal_config.h"
```
randomstuff: In fact we don't need `internal_config.h` but we need `simgrid_config.h` for `HAVE_MC`.

mquinson:
HAVE_MC is also defined (or not) in simgrid_config.h We should take that include instead.

PR #78, opened 2016-05-17
[s4u] add static method to create new  actors.

adfaure:
Hi, 
In s4u, when we create an actor (with the c++ constructor), the `simcall_process_create` function is called inside the constructor. The method `main`  of an actor is a virtual function of actor.  This method is called into `simcall_process_create` before it returns. It lead to a case where a virtual function is called inside of the constructor of a derived class. The results is that the method  `main` called when we create a process is the method of the base class.  
This pull request adds a static function `createActor` to create new actor. This method creates an object of actor in the first place and then calls `simcall_process_create`. 
The limitation is that the new actor class must **publicly** inherit from the actor.

### DIFF
--- a/include/simgrid/s4u/actor.hpp
+++ b/include/simgrid/s4u/actor.hpp
@@ -54,14 +54,14 @@ public:
   
   /** Method to create a new actor, the template parameter must extend the class Actor. In addition  the inheritance must be public. */
   template<class ActorSubClass> static ActorSubClass *createActor(const char*name, s4u::Host *host, int argc, char **argv, double killTime) {
-     Actor *res = dynamic_cast<Actor*>(new ActorSubClass());
+     ActorSubClass *res = new ActorSubClass();
      xbt_assert(res,"The actor class must extends the class simgrid::s4u::Actor.");
      
-     res->pimpl_ = simcall_process_create(name, s4u_actor_runner, static_cast<void*>(res), host->name().c_str(), killTime, argc, argv, NULL/*properties*/,1);
+     static_cast<Actor*>(res)->pimpl_ = simcall_process_create(name, s4u_actor_runner, static_cast<void*>(res), host->name().c_str(), killTime, argc, argv, NULL/*properties*/,1);
      xbt_assert(res->pimpl_,"Cannot create the actor");
    //  TRACE_msg_process_create(procname, simcall_process_get_PID(p_smx_process), host->getInferior());
    //  simcall_process_on_exit(p_smx_process,(int_f_pvoid_pvoid_t)TRACE_msg_process_kill,p_smx_process);
-     return dynamic_cast<ActorSubClass*>(res);
+     return res;
   };
   
   Actor();
--- a/include/simgrid/s4u/actor.hpp
+++ b/include/simgrid/s4u/actor.hpp
@@ -10,8 +10,11 @@
 #include <simgrid/simix.h>
 #include <simgrid/s4u/forward.hpp>
 
+int s4u_actor_runner(int argc, char **argv);
+
 namespace simgrid {
 namespace s4u {
+
 
 /** @brief Simulation Agent
  *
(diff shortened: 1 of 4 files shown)


include/simgrid/s4u/actor.hpp
```diff
@@ -39,10 +42,29 @@ namespace s4u {
...
-  Actor(const char*name, s4u::Host *host, int argc, char **argv, double killTime);
+  
+  /** Method to create a new actor, the template parameter must extend the class Actor. In addition  the inheritance must be public.  */
+  template<class ActorSubClass> static ActorSubClass *createActor(const char*name, s4u::Host *host, int argc, char **argv) {
+    return createActor<ActorSubClass>( name, host, argc, argv, -1);
+  };
+  
+  /** Method to create a new actor, the template parameter must extend the class Actor. In addition  the inheritance must be public. */
+  template<class ActorSubClass> static ActorSubClass *createActor(const char*name, s4u::Host *host, int argc, char **argv, double killTime) {
+     Actor *res = dynamic_cast<Actor*>(new ActorSubClass());
```
randomstuff: We can just use `ActorSubClass*` and avoid all the casts here.

We can use a static assert such as:

    static_assert(std::is_base_of<Actor, ActorSubClass>::value, "Actor::createActor only works with subclasses of Actor");

in order to enforce the types.

include/simgrid/s4u/actor.hpp
```diff
@@ -39,10 +42,29 @@ namespace s4u {
...
-  Actor(const char*name, s4u::Host *host, int argc, char **argv);
-  Actor(const char*name, s4u::Host *host, int argc, char **argv, double killTime);
+  
+  /** Method to create a new actor, the template parameter must extend the class Actor. In addition  the inheritance must be public.  */
+  template<class ActorSubClass> static ActorSubClass *createActor(const char*name, s4u::Host *host, int argc, char **argv) {
+    return createActor<ActorSubClass>( name, host, argc, argv, -1);
+  };
+  
+  /** Method to create a new actor, the template parameter must extend the class Actor. In addition  the inheritance must be public. */
+  template<class ActorSubClass> static ActorSubClass *createActor(const char*name, s4u::Host *host, int argc, char **argv, double killTime) {
```
randomstuff: Do we need some sort of smart pointer in order to correctly manage the lifetime of the Actor?

adfaure:
Yes, that's how I did it in the first place however `pimpl_` is a private class member hence we cannot access it if we don't cast it into an Actor.

include/simgrid/s4u/actor.hpp
adfaure: Yes, that's how I did it in the first place however `pimpl_` is a private class member hence we cannot access it if we don't cast it into a base class.

randomstuff:
As we discussed (for reference), I tend to think we should have a clear separation (not in the same class) between:

1. the API used to manipulate the actor (eg. `kill`);

2.  the API used by the actor (eg. `sleep`) and the code of the actor.

See `std::thread` where `std::thread` is used to manipulate the thread, the `std::this_thread` (external API) namespace is used by the thread (internal API) and the code of the thread is passed as an object to the constructor of the (non-plymorphic) `std::thread` class.

randomstuff:
Seems to be working allright:

    // This is a simplified version :)
    class Actor {
    private:
      int pimpl;
    public:
      template<class X>
      static X* create()
      {
        X* x = new X();
        x->pimpl = 42;
        return x;
      }
    };
    
    class MyActor : public Actor {};
    
    int main()
    {
      Actor::create<MyActor>();
      return 0;
    }

I guess, the best way is to:

    static_cast<Actor*>(res)->pimpl_ = simcall_process_create(name, s4u_actor_runner, static_cast<void*>(res), host->name().c_str(), killTime, argc, argv, NULL/*properties*/,1);

In fact, my main (implicit) complain was that we don't need a dynamic cast but only a static one in this context (we are creating a `ActorSubClass` and upcasting it to `Actor`).

include/simgrid/s4u/actor.hpp
```diff
@@ -54,14 +54,14 @@ XBT_PUBLIC_CLASS Actor {
   
   /** Method to create a new actor, the template parameter must extend the class Actor. In addition  the inheritance must be public. */
   template<class ActorSubClass> static ActorSubClass *createActor(const char*name, s4u::Host *host, int argc, char **argv, double killTime) {
-     Actor *res = dynamic_cast<Actor*>(new ActorSubClass());
+     ActorSubClass *res = new ActorSubClass();
      xbt_assert(res,"The actor class must extends the class simgrid::s4u::Actor.");
      
-     res->pimpl_ = simcall_process_create(name, s4u_actor_runner, static_cast<void*>(res), host->name().c_str(), killTime, argc, argv, NULL/*properties*/,1);
+     static_cast<Actor*>(res)->pimpl_ = simcall_process_create(name, s4u_actor_runner, static_cast<void*>(res), host->name().c_str(), killTime, argc, argv, NULL/*properties*/,1);
```
randomstuff: In fact, I thunk there's still a (racy?) problem here because when the actor starts executing, the `pimpl_` field might still be `nullptr`.

randomstuff:
I guess we should make the constructor of `Actor`, `protected` in this design.

mquinson:
I love the idea of splitting the public and internal APIs.

randomstuff:
I'm working in this direction (move the internal API out of the Actor class) as a replacement for this PR.

PR #82, opened 2016-06-06
[S4U] add kill actor with PID

adfaure:
Hi,  this commit add a way to kill an actor from its PID.
I needed it in RSG, so I did it.

### DIFF
--- a/src/s4u/s4u_actor.cpp
+++ b/src/s4u/s4u_actor.cpp
@@ -53,6 +53,17 @@ double s4u::Actor::getKillTime() {
   return simcall_process_get_kill_time(pimpl_);
 }
 
+void s4u::Actor::kill(int PID) {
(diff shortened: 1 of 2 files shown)


src/s4u/s4u_actor.cpp
```diff
@@ -53,6 +53,15 @@ double s4u::Actor::getKillTime() {
   return simcall_process_get_kill_time(pimpl_);
 }
 
+void s4u::Actor::kill(int PID) {
+  msg_process_t process = SIMIX_process_from_PID(PID);
+  if(process != NULL) {
+    simcall_process_kill(process);
+  } else {
+    // maybe raise exeption or xbt_die ?
```
randomstuff: Yes, we should throw an exception in this case.

PR #84, opened 2016-06-10
[s4u] add mutex

adfaure:
An example can be found at  `examples/s4u/actor/`.

### DIFF
--- a/include/simgrid/s4u/mutex.hpp
+++ b/include/simgrid/s4u/mutex.hpp
@@ -1,0 +1,36 @@
+/* Copyright (c) 2006-2015. The SimGrid Team. All rights reserved.          */
+
+/* This program is free software; you can redistribute it and/or modify it
+ * under the terms of the license (GNU LGPL) which comes with this package. */
+
+#ifndef SIMGRID_S4U_MUTEX_HPP
+#define SIMGRID_S4U_MUTEX_HPP
+
+#include <xbt/base.h>
+#include "simgrid/simix.h"
+
+
+namespace simgrid {
+namespace s4u {
+
+XBT_PUBLIC_CLASS Mutex {
+
+public:
+  Mutex();
+  ~Mutex() {};
+
+protected:
+
+public:
+
+  void lock();
+  void unlock();
+  bool try_lock();
+
+private:
+  std::shared_ptr<s_smx_mutex> _mutex;
+
+};
+}} // namespace simgrid::s4u
+
+#endif /* SIMGRID_S4U_MUTEX_HPP */
--- a/src/s4u/s4u_mutex.cpp
+++ b/src/s4u/s4u_mutex.cpp
@@ -1,0 +1,31 @@
+/* Copyright (c) 2006-2015. The SimGrid Team.
+ * All rights reserved.                                                     */
+
+/* This program is free software; you can redistribute it and/or modify it
+ * under the terms of the license (GNU LGPL) which comes with this package. */
+
+#include "xbt/log.h"
+#include "src/msg/msg_private.h"
+#include "src/simix/smx_network_private.h"
+
+#include "simgrid/s4u/mutex.hpp"
+
+
+using namespace simgrid;
+
+s4u::Mutex::Mutex() {
+    smx_mutex_t smx_mutex = simcall_mutex_init();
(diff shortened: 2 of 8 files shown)


include/simgrid/s4u/mutex.hpp
```diff
@@ -0,0 +1,47 @@
...
+ * Rendez-vous point for network communications, similar to URLs on which you could post and retrieve data.
+ * They are not network locations: you can post and retrieve on a given mailbox from anywhere on the network.
+ * You can access any mailbox without any latency. The network delay are only related to the location of the
+ * sender and receiver.
+ */
+XBT_PUBLIC_CLASS Mutex {
+
+public:
+  Mutex();
+  ~Mutex() {};
```
randomstuff: We should destroy `_mutex` here.

include/simgrid/s4u/mutex.hpp
```diff
@@ -0,0 +1,47 @@
...
+namespace s4u {
+
+/** @brief Mailboxes
+ *
+ * Rendez-vous point for network communications, similar to URLs on which you could post and retrieve data.
+ * They are not network locations: you can post and retrieve on a given mailbox from anywhere on the network.
+ * You can access any mailbox without any latency. The network delay are only related to the location of the
+ * sender and receiver.
+ */
+XBT_PUBLIC_CLASS Mutex {
```
randomstuff: If `Mutex` implements unique onwership of the underlying mutex it should not be copyable but moveable.

randomstuff: Maybe we should try to follow the [Mutex](http://en.cppreference.com/w/cpp/concept/Mutex) concept: non copyable, non moveable.

randomstuff: In fact, we only need the `Lockable` concept to make it work with `lock_guard` and other.

include/simgrid/s4u/mutex.hpp
```diff
@@ -0,0 +1,46 @@
...
+#define SIMGRID_S4U_MUTEX_HPP
+
+#include <xbt/base.h>
+#include "simgrid/simix.h" 
+
+
+namespace simgrid {
+namespace s4u {
+
+/** @brief Mailboxes
```
randomstuff: This one is wrong :)

src/s4u/s4u_mutex.cpp
```diff
@@ -14,17 +14,18 @@
 using namespace simgrid;
 
 s4u::Mutex::Mutex() {
-    _mutex = simcall_mutex_init();  
+    smx_mutex_t smx_mutex = simcall_mutex_init();  
+    _mutex = std::shared_ptr<s_smx_mutex>(smx_mutex, [](smx_mutex_t mutex){ SIMIX_mutex_destroy(mutex); });
```
randomstuff: In fact, we can do this:

    _mutex = std::shared_ptr<s_smx_mutex>(smx_mutex, SIMIX_mutex_destroy);

which is probably simplet :)

examples/s4u/actor/s4u_actor.cpp
```diff
@@ -0,0 +1,87 @@
...
+  }
+};
+
+class MainActor {
+public:
+  void operator()() {
+    int *res = new int;
+    simgrid::s4u::Mutex *mtx = new simgrid::s4u::Mutex();
+    *res = 0;
+    simgrid::s4u::Actor* workers[NB_ACTOR*2];
```
randomstuff: I get this warning:

    simgrid/examples/s4u/actor/s4u_actor.cpp:68:26: error: variable ‘workers’ set but not used [-Werror=unused-but-set-variable]

examples/s4u/actor/s4u_actor.cpp
```diff
@@ -0,0 +1,87 @@
+/* Copyright (c) 2006-2015. The SimGrid Team. All rights reserved.          */
```
randomstuff: The name of this test should probably be changed to include "mutex" inside it if the tests is supposed to be about mutexes.

examples/s4u/actor/s4u_actor.cpp
adfaure: Yes, initially I wanted to join those actors.

randomstuff:
Merged (with a small fix).

PR #85, opened 2016-06-16
[simix] Refcounting with SIMIX_cond_{ref,unref}()

adfaure:


src/simix/smx_synchro.cpp
```diff
@@ -356,19 +358,34 @@ void SIMIX_cond_broadcast(smx_cond_t cond)
...
   XBT_DEBUG("Destroy condition %p", cond);
-
   if (cond != nullptr) {
+    intrusive_ptr_release(cond);
+  }
+  XBT_OUT();
+}
+
+
+void intrusive_ptr_add_ref(s_smx_cond_t *cond)
```
randomstuff: I would define those `inline` in the header I guess.

randomstuff:
Looks nice. I'll compile/test that tomorrow before merging it.

src/simix/smx_synchro.cpp
adfaure: I would to but the header in which s_smx_cond_t is defined is private which lead to undefined references. Maybe could I put it into simix.h ?

randomstuff:
Merged. Thanks!

PR #86, opened 2016-06-17
[s4u] add condition variables

adfaure:
An example can be found at `examples/s4u/actions-comm`

### DIFF
--- a/include/simgrid/s4u/conditionVariable.hpp
+++ b/include/simgrid/s4u/conditionVariable.hpp
@@ -1,0 +1,67 @@
+/* Copyright (c) 2006-2016. The SimGrid Team. All rights reserved.          */
+
+/* This program is free software; you can redistribute it and/or modify it
+ * under the terms of the license (GNU LGPL) which comes with this package. */
+
+#ifndef SIMGRID_S4U_COND_VARIABLE_HPP
+#define SIMGRID_S4U_COND_VARIABLE_HPP
+
+#include <utility> // std::swap
+
+#include <simgrid/simix.h>
+#include <simgrid/s4u/mutex.hpp>
+
+namespace simgrid {
+namespace s4u {
+
+class Mutex;
+
+XBT_PUBLIC_CLASS ConditionVariable {
+  
+public:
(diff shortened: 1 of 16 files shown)


src/s4u/s4u_conditionVariable.cpp
```diff
@@ -0,0 +1,36 @@
+#include "simgrid/s4u/conditionVariable.hpp"
+#include "simgrid/simix.h"
+
+using namespace simgrid;
+
+s4u::CondVariable::CondVariable()  : cond_(simcall_cond_init()){
+    
+}
+
+s4u::CondVariable::~CondVariable() {
```
randomstuff: You need to [manage the references](https://github.com/adfaure/simgrid/blob/68111906f834c77ebf37048344d4970dcc44c7cf/include/simgrid/s4u/mutex.hpp#L31) when colpying/moving the object. Otherwise, the refcount will be broken in presence of copies (and result in dangling pointers).

include/simgrid/s4u/conditionVariable.hpp
```diff
@@ -0,0 +1,41 @@
...
+
+namespace simgrid {
+namespace s4u {
+
+class Mutex;
+
+XBT_PUBLIC_CLASS CondVariable {
+  
+public:
+CondVariable();
```
randomstuff: I'd add the constructor taking a `smx_cond_t` (add adding a reference) in order to be able to create `Mutex` from existing `mutex`.

include/simgrid/s4u/conditionVariable.hpp
```diff
@@ -0,0 +1,41 @@
...
+
+#include <simgrid/s4u/mutex.hpp>
+
+
+namespace simgrid {
+namespace s4u {
+
+class Mutex;
+
+XBT_PUBLIC_CLASS CondVariable {
```
randomstuff: I'd be inclined to call this `ConditionVariable`.

randomstuff:
This PR was merged but is not showed as merged because I merged a previous iteration (I missed an update).

PR #99, opened 2016-07-21
[SIMIX]add timeout to simcall_comm_waitany

adfaure:


src/simix/smx_network.cpp
```diff
@@ -476,12 +476,14 @@ void simcall_HANDLER_comm_testany(smx_simcall_t simcall, xbt_dynar_t synchros)
...
 }
 
-void simcall_HANDLER_comm_waitany(smx_simcall_t simcall, xbt_dynar_t synchros)
+void simcall_HANDLER_comm_waitany(smx_simcall_t simcall, xbt_dynar_t synchros, double timeout)
 {
   smx_synchro_t synchro;
   unsigned int cursor = 0;
 
   if (MC_is_active() || MC_record_replay_is_active()){
+    if(timeout != -1)
```
randomstuff: For consistency with the rest of the code, you might want to add space bewteen the `if` and the condition

src/simix/smx_network.cpp
```diff
@@ -490,7 +492,17 @@ void simcall_HANDLER_comm_waitany(smx_simcall_t simcall, xbt_dynar_t synchros)
     SIMIX_comm_finish(synchro);
     return;
   }
-
+  
+  if(timeout == -1 ){
```
randomstuff: Same.

src/simix/smx_network.cpp
```diff
@@ -490,7 +492,17 @@ void simcall_HANDLER_comm_waitany(smx_simcall_t simcall, xbt_dynar_t synchros)
...
+    simcall->timer = NULL;
```
randomstuff: We can use `nullptr`.

src/simix/smx_network.cpp
```diff
@@ -582,6 +594,8 @@ void SIMIX_comm_finish(smx_synchro_t synchro)
       continue; // if process handling comm is killed
     if (simcall->call == SIMCALL_COMM_WAITANY) {
       SIMIX_waitany_remove_simcall_from_actions(simcall);
+      if(simcall->timer)
```
randomstuff: Same.

randomstuff:
Looks good. I'd wait for the rest of the code before merging in order to know it actually works as intended :)

src/simix/smx_network.cpp
```diff
@@ -582,6 +594,8 @@ void SIMIX_comm_finish(smx_synchro_t synchro)
       continue; // if process handling comm is killed
     if (simcall->call == SIMCALL_COMM_WAITANY) {
       SIMIX_waitany_remove_simcall_from_actions(simcall);
+      if (simcall->timer) 
+        SIMIX_timer_remove(simcall->timer);
```
randomstuff: I think we should set `simcall->timer` to `nullptr` here in order to avoid letting some dangling pointer.

src/s4u/s4u_comm.cpp
```diff
@@ -18,6 +18,17 @@ Comm::~Comm() {
...
 
+template<class I>I wait_any(I first, I last, double timeout) {
+  xbt_dynar_t comms;           /* of msg_comm_t */
+  for(I iter = first; iter != last; iter++) {
+    xbt_dynar_push_back(comms, iter->pimpl_);
+  }
+  int idx = simcall_comm_waitany(comms, timeout);
+  if (idx == -1)
+    return last;
+  return first + idx;
```
randomstuff: We can use `std::next()` here which works with non-random-access iterators.

PR #100, opened 2016-07-22
Add a return value for MSG_sem_get_capacity

dindon-sournois:
There were no way of getting the value of the semaphore with MSG_sem_get_capacity.


mquinson:
Thanks for fixing this obvious copy/paste error

mquinson:
You could come and discuss with us on irc if you want: http://simgrid.gforge.inria.fr/simgrid/3.13.90/doc/community_contact.html

PR #114, opened 2016-09-12
fix segmentation fault when process try to unlock a not locked mutex

adfaure:


mquinson:
Thanks

src/simix/smx_synchro.cpp
```diff
@@ -163,6 +163,8 @@ bool Mutex::try_lock(smx_actor_t issuer)
 void Mutex::unlock(smx_actor_t issuer)
 {
   XBT_IN("(%p, %p)", this, issuer);
+  if(!this->locked)
+    THROWF(mismatch_error, 0, "Cannot release that mutex: it was not locked.");
```
randomstuff: I'd tend to think it'd be best to avoid using "XBT C-style exceptions" in new code.

mquinson: Woops I was really too fast to click on the merge button, sorry. You're right, C exception must die (or at least not multiply).
Thx, Mt.

Le 13 septembre 2016 00:08:50 GMT+02:00, Gabriel Corona <notifications@github.com> a écrit :
>> @@ -163,6 +163,8 @@ bool Mutex::try_lock(smx_actor_t issuer)
>>  void Mutex::unlock(smx_actor_t issuer)
>>  {
>>    XBT_IN("(%p, %p)", this, issuer);
>> +  if(!this->locked)
>> +    THROWF(mismatch_error, 0, "Cannot release that mutex: it was not
>locked.");
>
>I'd tend to think it'd be best to avoid using "XBT C-style exceptions"
>in new code.
>
>-- 
>You are receiving this because you modified the open/close state.
>Reply to this email directly or view it on GitHub:
>https://github.com/simgrid/simgrid/pull/114/files/7e9aea9de18ab183ef5e12261f92c3997d607dea#r78464309

-- 
Envoyé de mon téléphone. Veuillez excuser ma brièveté.

PR #124, opened 2016-11-30
[Energy platform parsing] assert + memory leak

mpoquet:
Hello,

When I updated SimGrid and tried to use my simulator with old energy platform files
(watts are defined as "min:max" instead of "idle:min:max"), I had some obscure OOB error.

I updated the assertion about the number of ":" subparts and
realized one string was not freed, so I fixed it too.

Best regards,
Millian


mquinson:
Thanks! This code was obviously written in a rush before an article. I seem to remember that the documentation is still to be updated, though.

PR #125, opened 2016-12-04
Simgrid for MPI load balancing simulation

rktesser:
As requested by Martin Quinson, I created this pull request to work on integrating into SimGrid the changes I did to to support my MPI load balancing simulation.


src/smpi/private.h
```diff
@@ -171,15 +171,21 @@ XBT_PRIVATE smx_mailbox_t smpi_process_mailbox();
...
 XBT_PRIVATE smx_mailbox_t smpi_process_remote_mailbox_small(int index);
+XBT_PRIVATE smx_mailbox_t smpi_process_remote_mailbox_migration(int index);
 XBT_PRIVATE xbt_mutex_t smpi_process_mailboxes_mutex();
 XBT_PRIVATE xbt_mutex_t smpi_process_remote_mailboxes_mutex(int index);
 XBT_PRIVATE xbt_os_timer_t smpi_process_timer();
 XBT_PRIVATE void smpi_process_simulated_start();
-XBT_PRIVATE double smpi_process_simulated_elapsed();
+
+/* 
+ * By rktesser: double smpi_process_simulated_elapsed() and
```
mquinson: This is a commit log, not a comment. Please remove this. (and I feel nervous about exposing stuff to the public API)

src/smpi/private.h
```diff
@@ -330,7 +336,11 @@ XBT_PRIVATE int smpi_mpi_waitall(int count, MPI_Request requests[], MPI_Status s
 XBT_PRIVATE int smpi_mpi_waitsome(int incount, MPI_Request requests[], int *indices, MPI_Status status[]);
 XBT_PRIVATE int smpi_mpi_testsome(int incount, MPI_Request requests[], int *indices, MPI_Status status[]);
 XBT_PRIVATE void smpi_mpi_bcast(void *buf, int count, MPI_Datatype datatype, int root, MPI_Comm comm);
-XBT_PRIVATE void smpi_mpi_barrier(MPI_Comm comm);
+
+/*
+ * By rktesser: smpi_mpi_barrier was moved to ../include/smpi/smpi_interface.h
```
mquinson: same here

src/smpi/private.h
```diff
@@ -414,8 +424,11 @@ XBT_PRIVATE void smpi_get_executable_global_size();
 XBT_PRIVATE void smpi_initialize_global_memory_segments();
 XBT_PRIVATE void smpi_destroy_global_memory_segments();
 XBT_PRIVATE void smpi_bench_destroy();
-XBT_PRIVATE void smpi_bench_begin();
-XBT_PRIVATE void smpi_bench_end();
+
+/* 
+ * By rktesser: smpi_bench_begin() and smpi_bench_end() where moved to
```
mquinson: same here

src/smpi/private.h
```diff
@@ -708,7 +721,14 @@ XBT_PRIVATE void TRACE_smpi_send(int rank, int src, int dst, int tag, int size);
 XBT_PRIVATE void TRACE_smpi_recv(int rank, int src, int dst, int tag);
 XBT_PRIVATE void TRACE_smpi_init(int rank);
 XBT_PRIVATE void TRACE_smpi_finalize(int rank);
-XBT_PRIVATE char *smpi_container(int rank, char *container, int n);
+XBT_PRIVATE void TRACE_smpi_process_change_host(int rank, sg_host_t host, sg_host_t new_host, int size);
+XBT_PRIVATE void TRACE_smpi_send_process_data_in(int rank);
+XBT_PRIVATE void TRACE_smpi_send_process_data_out(int rank);
+
+/*
```
mquinson: same here

src/smpi/smpi_base.cpp
```diff
@@ -1434,3 +1434,49 @@ void smpi_mpi_exscan(void *sendbuf, void *recvbuf, int count, MPI_Datatype datat
...
+  smx_mailbox_t mailbox;
+  MPI_Request request = nullptr; /* MC needs the comm to be set to NULL during the call */
+  int dest_rank;
+  double sleep_time = 0;
+  xbt_swag_t proc_list = sg_host_get_process_list(dest);
+  smx_actor_t dest_proc = (smx_actor_t) xbt_swag_extract(proc_list);
+  xbt_swag_insert(dest_proc, proc_list);
+  dest_rank = smpi_process_index_of_smx_process(dest_proc);
+  mailbox = smpi_process_remote_mailbox_migration(dest_rank);
+//MPI_Request request = smpi_mpi_isend(nullptr, size, MPI_BYTE, dest_rank,
```
mquinson: dead code

src/smpi/smpi_base.cpp
```diff
@@ -1434,3 +1434,49 @@ void smpi_mpi_exscan(void *sendbuf, void *recvbuf, int count, MPI_Datatype datat
...
+//			0, MPI_COMM_WORLD);
+  
+  //This call is based on smpi_mpi_isend.
+  request =  build_request(nullptr, size, MPI_BYTE, smpi_process_index(),
+      smpi_group_index(smpi_comm_group(MPI_COMM_WORLD), dest_rank), 0,
+      MPI_COMM_WORLD, NON_PERSISTENT | ISEND | SEND);
+  
+  //FIXME I'm not sure about this sleep time thing.
```
mquinson: Then, sort it out :)

src/smpi/smpi_base.cpp
```diff
@@ -1434,3 +1434,49 @@ void smpi_mpi_exscan(void *sendbuf, void *recvbuf, int count, MPI_Datatype datat
...
+  sleep_time = smpi_ois(request->size);
+  if(sleep_time > 0){
+    simcall_process_sleep(sleep_time);
+  }
+
+  //Not sure abou this block either.
```
mquinson: What is your question exactly?

src/smpi/smpi_pmpi.cpp
```diff
@@ -907,16 +910,92 @@ int PMPI_Ssend_init(void* buf, int count, MPI_Datatype datatype, int dst, int ta
...
 }
 
-int PMPI_Start(MPI_Request * request)
+int persistent_send(MPI_Request *request)
 {
   int retval = 0;
+  MPI_Request req = *request;
 
-  smpi_bench_end();
+  //smpi_bench_end();
```
mquinson: dead code

src/smpi/smpi_pmpi.cpp
```diff
@@ -907,16 +910,92 @@ int PMPI_Ssend_init(void* buf, int count, MPI_Datatype datatype, int dst, int ta
...
   if (request == nullptr || *request == MPI_REQUEST_NULL) {
     retval = MPI_ERR_REQUEST;
   } else {
+
+    int rank = req->comm != MPI_COMM_NULL ? smpi_process_index() : -1;
```
mquinson: Please comment your code. What are you doing here?

include/simgrid/host.h
```diff
@@ -26,6 +27,8 @@ XBT_PUBLIC(void*) sg_host_extension_get(sg_host_t host, size_t rank);
 XBT_PUBLIC(sg_host_t) sg_host_by_name(const char *name);
 XBT_PUBLIC(const char*) sg_host_get_name(sg_host_t host);
 XBT_PUBLIC(xbt_dynar_t) sg_hosts_as_dynar();
+XBT_PUBLIC(xbt_swag_t) sg_host_get_process_list(sg_host_t host);
```
rktesser: This is a re-implementation of a function that was there when I started working with SimGrid, but was later removed. The only thing this version does is return host->processes.

If really needed, I can make a function that builds a dynar by iterating through the swag. The major drawback is that I'll need to create and destroy the dynar several times. I can't reuse it because process migrations change the process list.

src/smpi/smpi_pmpi.cpp
rktesser: This is function is based on PMPI_Isend (actually, most of the code was copied from it). This test is present in that function too.

include/simgrid/host.h
mquinson: Yes, you need that function building a dynar. Or even better if your caller code is in C++, build a std::vector.

mquinson: To make it a C++ function, then your function should be in s4u::Host

src/instr/instr_TI_trace.cpp
```diff
@@ -219,11 +236,49 @@ void print_TIPushState(paje_event_t event)
...
+    extra->print_pop(trace_file, process_id, extra);
+  
   if (extra->recvcounts != nullptr)
     xbt_free(extra->recvcounts);
   if (extra->sendcounts != nullptr)
     xbt_free(extra->sendcounts);
-  xbt_free(process_id);
   xbt_free(extra);
+
+  return;
```
mquinson: What's the point of this 'return' at the end of a function? I see the point, actually, but that's not part of the SimGrid coding style.

src/smpi/instr_smpi.cpp
```diff
@@ -195,19 +204,21 @@ void TRACE_smpi_init(int rank)
...
-  container_t father;
+  container_t father, me;
   if (TRACE_smpi_is_grouped()){
     father = PJ_container_get (SIMIX_host_self_get_name());
   }else{
     father = PJ_container_get_root ();
   }
   xbt_assert(father!=nullptr,
       "Could not find a parent for mpi rank %s at function %s", str, __FUNCTION__);
+  me = PJ_container_new(str, INSTR_SMPI, father);
```
mquinson: Please prefer declaring and affecting value on the same line, as in "container_t me = PJ_container...."

I know it's not the case everywhere in the code base, but that's the kind of fixes I do all the time so I'd prefer to have it already done here. Thanks.

src/smpi/instr_smpi.cpp
```diff
@@ -433,3 +450,113 @@ void TRACE_smpi_recv(int rank, int src, int dst, int tag)
...
+    return;
+  }
+
+  container_t father;
+  if (TRACE_smpi_is_grouped()){
+    father = PJ_container_get(sg_host_get_name(host));
+  }else{
+    father = PJ_container_get_root();
+  }
+  xbt_assert(father!=NULL,
```
mquinson: xbt_assert presents a whole stack of calls, so you dont' need to add a ```__FUNCTION__``` in addition, do you?

src/smpi/instr_smpi.cpp
```diff
@@ -433,3 +450,113 @@ void TRACE_smpi_recv(int rank, int src, int dst, int tag)
...
+	      "Could not find a parent for mpi rank %s at function %s", str,
+	      __FUNCTION__);
+  PJ_container_new(str, INSTR_SMPI, father);
+}
+
+long long int get_migration_counter()
+{
```
mquinson: I am surprised to see that such code actually works for you. This is grossly non reentrant, so you'll need to add a comment explaining which assumption you make on the caller's behavior. Who is the caller, and how you know that calls to this function do not intermix when they are several migrations at the same time.

src/smpi/smpi_global.cpp
```diff
@@ -96,6 +97,12 @@ static char *get_mailbox_name_small(char *str, int index)
   return str;
 }
 
+static char *get_mailbox_name_migration(char *str, int index)
+{
+  snprintf(str, MAILBOX_NAME_MAXLEN, "mig-%0*x", static_cast<int> (sizeof(int) * 2), index);
```
mquinson: Does this really have to be that complicated? Why not a "mig-%d" ?

src/smpi/smpi_replay.cpp
```diff
@@ -382,24 +476,35 @@ static void action_test(const char *const *action){
     flag = smpi_mpi_test(&request, &status);
 
     XBT_DEBUG("MPI_Test result: %d", flag);
-    /* push back request in vector to be caught by a subsequent wait. if the test did succeed, the request is now nullptr.*/
-    get_reqq_self()->push_back(request);
 
+    //We can't call register_request, because request may be NULL.
+    //FIXME? Is this really needed? Does xbt_dict_t allow you to inser a null
```
mquinson: yes, putting null values is OK. That's a bit difficult to retrieve it (and make the difference between not finding and finding the null value), but that's possible.

src/smpi/smpi_replay.cpp
```diff
@@ -382,24 +476,35 @@ static void action_test(const char *const *action){
...
+    //pointer?
+    register_request_with_key(request, key);
+    //FIXME free the key?
+    //free(key);
```
mquinson: Yes, please. Valgrind will tell you what to clean...

src/smpi/smpi_replay.cpp
```diff
@@ -942,18 +1105,70 @@ static void action_allToAllv(const char *const *action) {
...
+  comm_amount[3] = 0;
+  action = simcall_execution_parallel_start("data_migration", 2, host_list, 
+	    comp_amount, comm_amount, 1.0, -1.0, 0);
+  simcall_execution_wait(action);
+
+  TRACE_smpi_send_process_data_out(smpi_process_index());
+}
+
+
+/* Based on MSG_process_migrate [src/msg/msg_process.c] */
```
mquinson: Argh, don't duplicate code, please!

src/smpi/smpi_replay.cpp
```diff
@@ -942,18 +1105,70 @@ static void action_allToAllv(const char *const *action) {
   xbt_free(recvdisps);
 }
 
-void smpi_replay_run(int *argc, char***argv){
+
+/*In previous versions, we called smpi_send_process_data (implemented on
```
mquinson: Ouch, don't duplicate so much code, that's harmful to the maintainance.

src/smpi/smpi_replay.cpp
```diff
@@ -998,44 +1223,76 @@ void smpi_replay_run(int *argc, char***argv){
...
 
+/*
+ * This function only initializes smpi_replay. It does not start the replay.
+ * You should call this one if you want to be able to register new actions
+ * from your code.
+ * To really start the replay you must call smpi_replay_run.
+ */
+void smpi_replay_init(int *argc, char ***argv)
+{
+  smpi_replay_initialize(argc, argv);
```
mquinson: Why do you need this function? Why don't you call smpi_replay_initialize() instead of smpi_replay_init() ?

src/smpi/smpi_replay.cpp
```diff
@@ -998,44 +1223,76 @@ void smpi_replay_run(int *argc, char***argv){
...
+ * original SimGrid master branch.
+ */
+void smpi_replay_start(int *argc, char ***argv)
+{
+  xbt_replay_action_runner(*argc, *argv);
+ 
   /* and now, finalize everything */
   double sim_time= 1.;
   /* One active process will stop. Decrease the counter*/
-  XBT_DEBUG("There are %zu elements in reqq[*]", get_reqq_self()->size());
```
mquinson: You should not convert from std:: to XBT without a __really__ good reason. What is you reason here?

rktesser: My code is actually older than your conversion to std::. When I wrote this code this variable was a xbt_dynar_t, which I replaced with a xbt_dict_t, so I could match requests inside MPI_Wait.

But, I will look to see if there is a dictionary type I can use in std::. This may take some time, since I know very little C++.

rktesser:
Yes, these modifications are to support a  load balancing simulator for MPI applications. This work is being advised by Arnaud.

Initially, I implemented everything inside SimGrid, but Arnaud requested that I implement it as a separated module, which just links to SimGrid. That's why I need the public symbols.

Here's the repo with the code of the load balancing simulation (the most current branch is called 'new_simgrid'): 
https://bitbucket.org/rktesser/smpi-lb

Be advised that this code is kind of a mess right now and there's lots of testing/debugging stuff in it, sorry.

src/smpi/instr_smpi.cpp
rktesser: OK. I'll remove it, as this seems not to be needed anymore.
Maybe this was not the case when I wrote this code, about two years ago?

src/smpi/instr_smpi.cpp
rktesser: Thanks for pointing this out. This used to work when I had both data migration and process migration inside the same call. Now that I split them and made data migration parallel, this has indeed become broken.

This was used to create a link in the traces, from the origin to the destination of the migration. But I am not actually using this information anymore. So, for now, I will simply remove this from the code.

src/smpi/instr_smpi.cpp
rktesser: Done!

adegomme:
Hi, 
your PR also triggers some warnings, which our paranoid continuous integration settings will fail. Can you please use the enable_compile_warnings cmake flag when compiling ?

For instance I can see
[  0%] Building CXX object CMakeFiles/simgrid.dir/src/smpi/instr_smpi.cpp.o
src/smpi/instr_smpi.cpp: In function ‘void TRACE_smpi_init(int)’:
src/smpi/instr_smpi.cpp:218:9: error: unused variable ‘val’ [-Werror=unused-variable]
   val_t val = PJ_value_new ("migrate", "0 0 0", eventype);
         ^
src/smpi/instr_smpi.cpp: In function ‘void TRACE_smpi_task_migrate(int, sg_host_t)’:
src/smpi/instr_smpi.cpp:456:6: error: no previous declaration for ‘void TRACE_smpi_task_migrate(int, sg_host_t)’ [-Werror=missing-declarations]
 void TRACE_smpi_task_migrate(int rank, sg_host_t host)
      ^
cc1plus: all warnings being treated as errors

src/smpi/smpi_global.cpp
rktesser: This was done like this to use the same naming scheme of the other two already existing mailboxes.

src/smpi/smpi_replay.cpp
rktesser: Sorry. To which code are you referring here?

src/smpi/smpi_replay.cpp
rktesser: What do you suggest here, then?
Is this really a duplication? I don't think so.

rktesser:
@adegomme It should compile now, with these last five commits I made.

rktesser:
@adegomme OK. Now it should give no warnings, even with compiler optimizations.

src/smpi/smpi_replay.cpp
schnorr: Perhaps @rktesser could exhaustively look for all duplicated code.

src/smpi/smpi_replay.cpp
schnorr: @rktesser, just sticking to `std::map` with the correct template should be enough, right?

rktesser: @schnorr This specific change has already been done. The structure was replaced by a std::unordered_map.

For some reason, GitHub did not detect the modification.

mquinson:
Hello, sorry for the delay, I was busy then in vacations then busy...

As for your pull request, I think we should first discuss of what you want to achieve with it. I will thus commit a bunch of bug reports to your bitbucklet repo to get it cleaned out, so that you need less things in this pull request, so that it becomes acceptable.

Thanks for your patience,
Mt.

mquinson:
Damn, it seems I cannot open bugs on your repo. Could you add this feature for me please?

In the meanwhile, the first report is to remove the use (and implementation) of `smpilb_barrier_t` in https://bitbucket.org/rktesser/smpi-lb/src/ba3e124987e42773703a28f0316bff1954de5320/smpi_lb.c?at=master&fileviewer=file-view-default#smpi_lb.c-31 Instead, you should use `msg_bar_t`, defined in https://github.com/simgrid/simgrid/blob/master/include/simgrid/msg.h#L445 It is surprising to use a msg datatype here, but this is the way to go.

Second request, create your barrier from https://bitbucket.org/rktesser/smpi-lb/src/ba3e124987e42773703a28f0316bff1954de5320/replay.c?at=master&fileviewer=file-view-default#replay.c-13 This way, you will not need the whole complexity of https://bitbucket.org/rktesser/smpi-lb/src/ba3e124987e42773703a28f0316bff1954de5320/smpi_lb.c?at=master&fileviewer=file-view-default#smpi_lb.c-473

I will continue after your code improves on thse two first points.

Thanks for your patience,
Mt

rktesser:
@mquinson I have enabled the issue tracker.

As for the barrier, when I started implementing this code, I was told I should not call MSG functions from inside SMPI. Is that not the case anymore?

At first, I was calling smpi_mpi_barrier(), but it was buggy, so I wrote my own implementation. This bug seems to be fixed now. So, maybe I could try to go back to use smpi_mpi_barrier()?

mquinson:
Yes, indeed, this is not the case anymore. You can use the MSG_bar_t. Actually, if you grep a bit, you'll see that this is already used from within SMPI. It seems preferable to smpi_mpi_barrier which will fire a lot of MPI messages. I'd say that this is not what you want here.

You should proceed and implement the 3 changes I requested on your bitbucklet repo, please.

Thx, Mt.

rktesser:
@mquinson 

As per your first request, I wrote a version of the code using msg_barrier_t.

As for your second request, creating the barrier at the start of the 'main' function does not solve the problem, as there are already multiple processes being simulated. I tested it, and each process created its own barrier and overwrote the contents of 'smplib_bar'.

Furthermore, even my complicated initialization code does not work there, as the MPI ranks are assigned only when I start the replay. Before that, all processes have the same rank (in my tests it was -333).

There's also the matter of the destruction of the barrier. So far, I was not able to find a way to do it without causing a segmentation fault.

What I need here is a barrier implementation in which either you don't need to initialize/destroy it (as done by MPI_Barrier) or every process can safely call the initialization and destruction functions.

I still need to look at your third request. But before that I need to do a merge, to update my code to the current revision of SimGrid.

rktesser:
@mquinson As it is now, `xbt_action register` does not allow to override the actions, because they would be replaced when `smpi_replay_run` is called. For my code to work without splitting the `smpi_replay_run` I will have to modify `xbt_action_register` so that it does not overwrite existing actions.

mquinson:
I will merge this manually. Please don't touch it.

mquinson:
Ok, I merged it, but  some of the tests fail. Could you please help me fixing them? I gave a little try, but I think you'll be much faster than me since you know your code.

I think we are almost here, and I'm willing to merge your branch as soon as it compiles & passes the tests.

Thanks for your patience,
Mt.

mquinson:
The breakage is visible here:  https://travis-ci.org/simgrid/simgrid/jobs/254678820#L7293

rktesser:
@mquinson 

Thanks for your help!
I will take a look on this tomorrow.

The failed tests are probably comparing the output traces with an expected output. This is probably the problem for the smpi-replay and ti-tracing tests, since my code generates a slightly different trace. For example, I register the tags on the communication calls, which was not done before.

Probably, the only straightforward way to fix this is to modify the tests themselves. Otherwise, I will need to modify the code to enable these modifications only when they are needed.

I have no idea what is happening with the call location code. I'll have to check what my code does that changes these traces.

The Java related error seems strange, as I haven't touch anything java related.

mquinson:
Indeed, the java test is randomly failling nowadays, and you should forget about it. Nothing related to your work.

If the trace format was extended, then yes, you should modify the test to take this into account. But I suspect that one of the tests is segfaulting. Maybe because of my merge.

BEWARE, I just changed my branch lb-merge to squash all your changes into a single commit. I'prefer to not have 100 commits with broken intermediate state. It makes git bisect much harder, and we sometimes need it. I would have prefered to have this commit authored by you, but I wanted to spare you the effort. If you want, go to master and do the following: 
```
git checkout simgrid-merge # branch with all your changes
git merge upstream-master # merge all changes coming from upstream
# fix the mess, make it compile, make it pass the tests
# Inspiration: https://github.com/simgrid/simgrid/commit/f1c4725bec89fa7f707b46d5f9b68542ab96439a

git checkout master # go to a branch that is even with the upstream master 
git diff -R simgrid-merge > /tmp/patch
git apply /tmp/patch
git commit -m "Branch #125 squashed into a single commit for sake of clarity" -a
```
and then make a new pull request with the resulting change, all in one file.

mquinson:
Could you please squash all your changes into a single commit?

rktesser:
I'll squash the commits as soon as I'm done. There are still some failing tests.
What I did up to now only fixes the smpi-replay tests. These two are still failing:
	192 - smpi-tracing-call-location (Failed)
	519 - tesh-smpi-replay-ti-tracing (Failed)

mquinson:
This is superseeded by #214, thus closing.

Issue #139, opened 2017-02-20
Allow pthread creation in SMPI


sthibaul:
StarPU creates new processes from SMPI. These should inherit privatization. This testcase:

[test-thread.c.txt](https://github.com/simgrid/simgrid/files/788173/test-thread.c.txt)

currently fails: it shows

[output.txt](https://github.com/simgrid/simgrid/files/788174/output.txt)

while it should show

[expected-output.txt](https://github.com/simgrid/simgrid/files/788175/expected-output.txt)

(i.e. global_rank should be 0 in the thread started from SMPI node 0).

Note: the mpi_thread_create and thread_create_wrapper are helpers that we are using to basically mimic what pthread_create() does in an MPI program. Perhaps simgrid could provide such kind of interface?

adegomme:
Hi,

I was failing to reproduce, until I noticed something odd in your outputs. Are you activating privatization ?
Because I don't see the line  [0.000000] [xbt_cfg/INFO] Configuration change: Set 'smpi/privatize-global-variables' to '1'

It seems to be working for me on this test, but I'm a bit puzzled and don't really know if this is accidental and would need another test or if it's just working (are we by chance supporting MPI_THREAD_MULTIPLE ?).

Regards,
Augustin

sthibaul:
Ooops, sorry, indeed. Also, the MSG_process_sleep() call should be dropped from thread_create_wrapper, it was a workaround that we were using in StarPU to avoid yet another privatization bug.

I'd be happy to see this testcase included in simgrid, though, so we are sure that the way we use threads does continue working :)

I wouldn't be too surprised that simgrid supports MPI_THREAD_MULTIPLE quite well since simgrid basically serializes almost everything. There might be some corner cases, though. All we need for StarPU is being able to run MPI_Wait in the thread, which should be easy to make work in a separate thread.

sthibaul:
Ah, there, with a more stressful test:

[test-thread.c.txt](https://github.com/simgrid/simgrid/files/796596/test-thread.c.txt)

to get the global variable directly from the beginning of the thread function. I'm getting:

[node-1.acme.org:wait recv:(3) 0.000000] [msg_test/INFO] new thread has rank 1 and 0
[node-0.acme.org:wait send:(4) 0.000000] [msg_test/INFO] new thread has rank 0 and 1

even if later one gets

[node-1.acme.org:wait recv:(3) 0.000000] [msg_test/INFO] 1 has rank 1 and 1
[...]

Issue #137, opened 2017-02-16
Wanted feature: global privatization through dynamic linker tricks


mquinson:
This bug is there to keep track of an ongoing discussion about a nice feature we are currently implementing. This text was edited to incorporate the input that others wrote as a comment below. Thanks to @sthibaul and @randomstuff for their valuable input (as usual).

The context is about the automatic global privatization when folding the MPI ranks as threads of the simulated processor. For now, right after the Unix process is loaded and setup, we make N copies of the process global segments, one copy per MPI rank. Then, on each context switch, we remap the area where the code searches its data onto the copy of the process we want to execute.

The idea is to avoid this segment remaping (for sake of performance). Instead, we want the system to load the binary N times. 

Writing our own library loader seems ways too complicated, but we could probably do it with a standard `dlopen()`, provided that the binary was compiled with `-fPIE` or `-fPIC` (Position Independent Executable/Code). `-fPIC` means that you recompile your code as a dynamic library, which is not optimal but could constitute a safe fallback implementation for windows. `-fPIE` created to allow the loader to change the position and order of the executable segments each time, to make it harder for security attacker to guess where things are. Nowadays, that option is enabled by default on Debian, so this is definitely something we can safely rely on: only applications doing horrible things with symbols have problems with it.

The main difficulty of this approach is that `ld.so` is too cleaver: it won't `dlopen` the same file several times. Only the first request is honored. `dlmopen` seems appealing to remove that limitation, but only 256 naming spaces can be done with this call. Plus, libsimgrid would also be duplicated which is obviously not exactly something we want. Instead, we need to work around this `ld.so` sanity check in order to `dlopen()` N times the same code. 

Patching `ld.so` could be doable since we only want to desactivate the test on the inode deciding to avoid the file loading. But still, even for such a simple change, modifying `ld.so` remains rather complex and hard to maintain. In addition, it is though to force our users to have an alternative `ld.so`. An alternative is to interpose `ld.so` is not possible with `LD_PRELOAD` since `ld.so` embeeds everything it needs as static objects, but you could still use `ptrace`. This really slows down the application, but you can disable it once the application is properly loaded. Still, it sounds rather difficult to achieve properly.

[...]

PR #140, opened 2017-02-24
SMPI privatization based on dlopen

randomstuff:
⚠️ **Not ready for merging.**

The idea is that the SMPI user code is now compiled as a dynamic library which is dynamically loaded into the Simgrid/SMPI process (the `smpi_main` executable):
* either once, for the classical privatization options ("no", "mmap");
* or once per simulated process, for the new privatization mode ("dlopen").

Basic stuff:

* [x] Repair classic privatization mode
* [x] Fix `smpi_main` handling (`PATH`, installation, etc.)
* [x] Make it work with Fortran (`mpiff`, `mpif90`)
* [x] Don't rely on external `cp` command
* [x] Cleanup

Cross platform:

* [ ] <del>Fix SMPI on Windows</del>
* [x] Fix SMPI on MacOS X
* [x] New privatization on BSD (probably won't work because of lack of `RTLD_DEEPBIND`)

Optional optimizations :

* [ ] Deduplicate in-memory pages of the text segment using `mmap` (unless a `reflink` was used)
* [ ] `ftruncate` and punch-holes in the on-disk copies

Debugging :

* [ ] Investigate behaviour wrt debuggers (unlinking the file immediately, additional messing-up with the file)

Model-checker:

* [ ] Fix the model checker
* [ ] Support for the new privatization mode


randomstuff:
Fortran fixed.

`100% tests passed, 0 tests failed out of 258` \o/

randomstuff:
The bad news is that for it to work correctly, we need `RTLD_DEEPBIND` which is GNU-libc specific.

mquinson:
RTLD_DEEPBIND is an ugly pimple to fix the behavior of the gnu ld. Other ld don't need it, so we're good.

From my POV, we still need to fix MC (if broken), and cleanup. Optimizations are not mandatory.

mquinson:
I confirm that MC is broken on that branch. Conditional jump or move depends on uninitialised value(s) on the following line of mc::~Process()
```C
  if (this->unw_underlying_addr_space != unw_local_addr_space) {
```

Ouch, too hard for me, I'm affraid.

mquinson:
Augustin told me that this branch is working on winbuntu, and SMPI never worked on Windows without that subsystem, so I think we're good on that front.

randomstuff:
As discussed in #139, reflink won't give us page sharing the page cache works per inode on Linux.

randomstuff:
See the [docker doc](https://docs.docker.com/engine/userguide/storagedriver/btrfs-driver/):

> **Page caching.** Btrfs does not support page cache sharing. This means that n containers accessing the same file require n copies to be cached. As a result, the btrfs driver may not be the best choice for PaaS and other high density container use cases.

randomstuff:
See alos [this thread](http://www.spinics.net/lists/linux-btrfs/msg38790.html) with a simple test showing the lack of page cache sharing between reflinks.

mquinson:
I implemented the file copy without the cp command.

mquinson:
So I'm a bit lost now. What are our options to deduplicate the pages in memory if neither bttr FS nor reflink will help us?

Is punchholing the file the only option we have? 

At the same time, I slowly get the feeling that this optimization is not needed in practice. In the case of HPL, the binary is only 800k, so that will definitely not be the performance killer.

Ezibenroc:
> At the same time, I slowly get the feeling that this optimization is not needed in practice. In the case of HPL, the binary is only 800k, so that will definitely not be the performance killer.

Using `dlopen` for privatization has been a great improvement in HPL.
But if I understood correctly, what you are working on now will only impact the initialization of the program, not the context switches. This initialization is already negligible in HPL, so I don't think it is usefull to look further.

mquinson:
Well, I'm not so sure, @Ezibenroc. What's killing your HPL right now seem to be related to memory usage. Ok, Mostly because the kernel fail to properly prepare the pages that are mmap'ed with the POPULATE flag, but still. I think that dedupplicating the code may help to grow above a given point. The next step seem to be the kernel patch prepared by @sthibaul, but it may well be that the move after this one is the dedupplication. Will see.

randomstuff:
@Ezibenroc, Deduplicating will reduce the memory consumption. Moreover we could expect better performance by reducing CPU cache usage (especially the instruction cache).

mquinson:
I improved the documentation of this feature, explaining what I understand from the work remaining to do. Could you guys please have a look? Thanks.

http://simgrid.gforge.inria.fr/simgrid/3.16/doc/options.html#options_smpi_global

Also, we will need a few test for this. We don't want to dupplicate the whole set of MPI tests, or they will last over one hour. @adegomme, could you please take care of this?

sthibaul:
The idea was, after dlopen()ing several times, to look in /proc/$pid/smaps and remap the original binary file over the file copies mapped by dlopen(), so that we do get code sharing, and only the dynamic linking bits are not shared. I.e. not reimplementing dlopen() to get the wanted behavior, but fix the result of dlopen().

mquinson:
Ok, we now have some tests of this feature, and it works like a charm on mac and freebsd. Simply, it does not like pthreads on these architectures: you gain a SIGBUS somewhere in malloc when you activate both dlopen and pthread at the same time.

But dlopen and raw contextes actually work on Mac and FreeBSD. Privatization is disabled on NetBSD and I did not try to force dlopen on that platform. The week-end was not *that* bad.

Next step will be the interaction with MC, I guess. And the code deduplication.

mquinson:
Closing this PR since there is nothing anymore to merge in it, and the remaining discussion took place in the associated bug #137.

PR #154, opened 2017-04-13
Partial shared malloc

Ezibenroc:
This implements the partial shared malloc and the no-memcpy optimization for communications with partial-shared-malloced blocks.

## User side

There is one new macro:
```c
/*
 * Allocate a contiguous block of memory of given size.
 * The sub-blocks described by shared_block_offsets are shared.
 * This array contains the offsets (in bytes) of the block to share.
 * Even indices are the start offsets (included), odd indices are the stop offsets (excluded).
 * For instance, if shared_block_offsets == {(27, 42), (100,200)}, then the elements mem[27], mem[28], ..., mem[41] and mem[100], ..., mem[199] are shared. The others are not.
 */
#define SMPI_PARTIAL_SHARED_MALLOC(size, shared_block_offsets, nb_shared_blocks)
```
For deallocation, simply use `SMPI_SHARED_FREE`.

## Programmer side

* Function `smpi_shared_malloc` has been splitted in two functions `smpi_shared_malloc_local` and `smpi_shared_malloc_global`.
* Function `smpi_shared_malloc_global` is a generalization of what was done for the global shared malloc. We iterate on all the shared blocks and `mmap` them by chunk on the bogus file.
* The struct `shared_metadata_t` has a new field, a vector of pairs, representing the private blocks of the given entry.
* This vector is provided to the caller of function `smpi_is_shared`. It is used in `smpi_comm_copy_buffer_callback`. Here, we iterate on the private blocks and call `memcpy` for each of them.
* There are two new tesh tests, one for the partial shared malloc and one for the communication optimization.

## Future work

* Handle the no-memcpy optimization for partial shared malloc in files `smpi_datatype.cpp` and `smpi_request.cpp`.
* Add unit tests for some functions (e.g. `shift_and_frame_private_blocks` and `merge_private_blocks`).


mquinson:
Whaou, nice work, many thanks. Could you please add the documentation part (the header of this pull request) to the SMPI tutorial in scm.gforge.inria.fr:/gitroot/simgrid/propaganda.git please ? Thanks in advance.

mquinson:
Hello,

would it be possible to reduce the time taken by the tests, somehow? It seems to take over 50 secondes when run under dynamic analysis with valgrind: https://ci.inria.fr/simgrid/view/Tous/job/SimGrid-DynamicAnalysis-Sanitizer-Thread/label=simgrid-debian8-64-dynamic-analysis-2/lastCompletedBuild/testReport/

If possible, it'd be cool if we could have the same amount of coverage in less time. We could also increase the timeout value, if needed for the testing quality.

Finally, please note that I did not investiguate at all. This may be the sign of a deeper problem found by the test. But I trust you to sort it out ;-)

Thanks for your help,
Mt

mquinson:
I just updated the SMPI tutorial to take the recent improvements into account: http://simgrid.gforge.inria.fr/tutorials/simgrid-smpi-101.pdf

@Ezibenroc, others, could you please have a look and report/fix any issue?
Thanks, Mt.

Ezibenroc:
> would it be possible to reduce the time taken by the tests, somehow?

A tentative here: b382315f0693083344eb9a25a8fc37b7091bddc7, I use sizes about 16 times lower (but still larger than the block size).
Could you tell me if it is enough? I tried the old version with valgrind, it was much quicker than these 50 seconds (but maybe there is some options to use?).

> I just updated the SMPI tutorial to take the recent improvements into account

I don't think the syntax `{(27,42),(100,200)}` is valid C. But it is maybe clearer than writing `{27,42,100,200}`.
Otherwise, it's great :+1:

mquinson:
The dynamic build will run this night. We'll see if it's enough.

For the doc, I just copyied what you have in comment in the introduction of this PR :) It's fixed now, thanks.

mquinson:
It was enough, indeed: https://ci.inria.fr/simgrid/view/Tous/job/SimGrid-DynamicAnalysis-Sanitizer-Thread/label=simgrid-debian8-64-dynamic-analysis-2/lastCompletedBuild/testReport/ your tests disappeard from the list of broken tests, which is good.

Thanks for your help,
Mt

PR #164, opened 2017-05-03
restore smpi_process_get_user_data and smpi_process_set_user_data

sthibaul:
This restores implementations of smpi_process_get_user_data and smpi_process_set_user_data


sthibaul:
Thanks!

PR #167, opened 2017-05-18
Add smpi_execute_public

Ezibenroc:
This function simply calls `smpi_execute`, but take care of stopping and restarting the bench.

It is useful to replace expensive pieces of application code that cannot be wrapped by a `SMPI_SAMPLE` because of their irregularity. For instance, in HPL, we use it to simulate `dgemm` and `dtrsm`.

Before that, we used `smpi_usleep`, but this one has two drawbacks:
- It takes a virtual time and not a real time, so you have to redo your linear regression each time you change the speed of the platform you simulate.
- It is a sleep, so it works for time estimation but not for energy.

Regarding the name of the function, I am not sure it is very explicit, but I had no better idea.

### DIFF
--- a/include/smpi/smpi.h
+++ b/include/smpi/smpi.h
@@ -843,6 +843,7 @@ XBT_PUBLIC(void) smpi_process_set_user_data(void *);
 
 XBT_PUBLIC(void) smpi_execute_flops(double flops);
 XBT_PUBLIC(void) smpi_execute(double duration);
+XBT_PUBLIC(void) smpi_execute_public(double duration);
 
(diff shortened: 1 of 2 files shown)


mquinson:
Could you please rename smpi_execute_public into smpi_execute_benched, please?

Thanks a lot,
Mt

Ezibenroc:
Sure, done in https://github.com/simgrid/simgrid/pull/167/commits/8c39f23c04ffdf69507a05539bb032a10571ec57

mquinson:
thanks a lot.

PR #168, opened 2017-05-18
Add support for huge pages in SMPI

Ezibenroc:
This adds an option to `smpirun` asking to use huge pages for the calls to `SMPI_SHARED_MALLOC` and `SMPI_PARTIAL_SHARED_MALLOC`.

To use it:
```bash
mkdir /home/huge && sudo mount none /home/huge -t hugetlbfs -o rw,mode=0777
sudo sh -c "/bin/echo 1 > /proc/sys/vm/nr_hugepages"
smpirun --cfg=smpi/shared-malloc-hugepage:/home/huge --cfg=smpi/shared-malloc-blocksize:2097152 [...]
```

For very large allocations, the page table grows very large and the kernel takes a lot of time working on this table, stalling the simulation.
For instance, without huge pages, HPL with a matrix of size 6e5 has a page table of 5.5GB and the simulation drops below 40% of CPU utilization.
Using huge pages fix this. Of course this only shifts the problem further, but this is an easy fix.

Currently, the size of a huge page is hard coded to be 2 MB, as I did not find a standard way to get it. Maybe we could just ask the user to provide it?

### DIFF
--- a/doc/doxygen/options.doc
+++ b/doc/doxygen/options.doc
@@ -1141,6 +1141,8 @@ at least one huge page:
     sudo sh -c 'echo 1 > /proc/sys/vm/nr_hugepages'
 \endcode
 
+Note that at least one huge page is required for smpirun, but you can set
+an higher number of huge pages in nr_hugepages (e.g. "echo 27").
 Then, you can pass the option --cfg=smpi/shared-malloc-hugepage:/home/huge
 to smpirun.
 
--- a/doc/doxygen/options.doc
+++ b/doc/doxygen/options.doc
@@ -1115,6 +1115,35 @@ only consume 1 MiB in memory.
 You can disable this behavior and come back to regular mallocs (for
 example for debugging purposes) using \c "no" as a value.
 
+If you want to keep private some parts of the buffer, for instance if these
+parts are used by the application logic and should not be corrupted, you
+can use SMPI_PARTIAL_SHARED_MALLOC(size, offsets, offsets_count).
+
+As an example,
+
+\code{.unparsed}
+    mem = SMPI_PARTIAL_SHARED_MALLOC(500, {27,42 , 100,200}, 2);
+\endcode
+
+will allocate 500 bytes to mem, such that mem[27..41] and mem[100..199]
+are shared and other area remain private.
+
+Then, it can be deallocated by calling SMPI_SHARED_FREE(mem).
+
+When smpi/shared-malloc:global is used, it is possible to optimize even
+further the memory consumption and the simulation time by using huge pages.
+To do so, it is required to mount a hugetlbfs on your system and allocate
+at least one huge page:
+
+\code{.unparsed}
+    mkdir /home/huge
+    sudo mount none /home/huge -t hugetlbfs -o rw,mode=0777
+    sudo sh -c 'echo 1 > /proc/sys/vm/nr_hugepages'
+\endcode
+
+Then, you can pass the option --cfg=smpi/shared-malloc-hugepage:/home/huge
+to smpirun.
+
 \subsection options_model_smpi_wtime smpi/wtime: Inject constant times for calls to MPI_Wtime
 
 \b Default value: 0
--- a/src/smpi/smpi_shared.cpp
+++ b/src/smpi/smpi_shared.cpp
@@ -224,14 +224,14 @@ static void *smpi_shared_malloc_local(size_t size, const char *file, int line)
 void* smpi_shared_malloc_partial(size_t size, size_t* shared_block_offsets, int nb_shared_blocks)
 {
   char *huge_page_mount_point = xbt_cfg_get_string("smpi/shared-malloc-hugepage");
-  const bool use_huge_page = huge_page_mount_point[0] != '\0';
-#ifndef MAP_HUGETLB
-    xbt_assert(!use_huge_page, "Huge pages are not available on your system.");
+  bool use_huge_page = huge_page_mount_point[0] != '\0';
+#ifndef MAP_HUGETLB /* If the system header don't define that mmap flag */
+    xbt_assert(!use_huge_page, "Huge pages are not available on your system, you cannot use the smpi/shared-malloc-hugepage option.");
+    use_huge_page = 0;
 #endif
   smpi_shared_malloc_blocksize = static_cast<unsigned long>(xbt_cfg_get_double("smpi/shared-malloc-blocksize"));
   void *mem, *allocated_ptr;
   size_t allocated_size;
-#ifdef MAP_HUGETLB
   if(use_huge_page) {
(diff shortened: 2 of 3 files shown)


mquinson:
Could you please add a test for it? It can be any existing test with the new flag passed, I guess.
If you don't want to mess with the cmake madness, just tell me the test you want to change and I'll take care of this.

Many thanks for this great feature!
Mt

src/smpi/smpi_shared.cpp
```diff
@@ -125,6 +127,7 @@ std::map<void*, shared_metadata_t> allocs_metadata;
 xbt_dict_t calls = nullptr;           /* Allocated on first use */
 #ifndef WIN32
 static int smpi_shared_malloc_bogusfile           = -1;
+static int smpi_shared_malloc_bogusfile_huge_page  = -1;
```
mquinson: Why do you need another global variable `smpi_shared_malloc_bogusfile_huge_page`? Can't you use  the `smpi_shared_malloc_bogusfile` variable in any case, regardless of whether we use huge pages or regular ones?

src/smpi/smpi_shared.cpp
```diff
@@ -218,23 +223,63 @@ static void *smpi_shared_malloc_local(size_t size, const char *file, int line)
  */
 void* smpi_shared_malloc_partial(size_t size, size_t* shared_block_offsets, int nb_shared_blocks)
 {
-  void *mem;
-  xbt_assert(smpi_shared_malloc_blocksize % PAGE_SIZE == 0, "The block size of shared malloc should be a multiple of the page size.");
+  char *huge_page_mount_point = xbt_cfg_get_string("smpi/shared-malloc-hugepage");
+  const bool use_huge_page = huge_page_mount_point[0] != '\0';
+#ifndef MAP_HUGETLB
```
mquinson: Please don't use #ifndef, because it makes it hard to detect whether you made a typo in the name (or removed a needed header file). Instead, use `#if !MAP_HUGETLB` or similar. With this, the `-Wundef` that we are using can detect the previous errors.

mquinson: Ok, I understand now that this is a standard mmap flag, defined in system headers. Sorry for the mistake. Please explain it with something like (to be used each time you have such a test):
```
 #ifndef MAP_HUGETLB /* If the system header don't define that mmap flag */
```

src/smpi/smpi_shared.cpp
```diff
@@ -218,23 +223,63 @@ static void *smpi_shared_malloc_local(size_t size, const char *file, int line)
...
+    xbt_assert(!use_huge_page, "Huge pages are not available on your system.");
```
mquinson: Be more explicit for the user: add "don't use the smpi/shared-malloc-hugepage option"

src/smpi/smpi_shared.cpp
```diff
@@ -218,23 +223,63 @@ static void *smpi_shared_malloc_local(size_t size, const char *file, int line)
...
+#endif
+  smpi_shared_malloc_blocksize = static_cast<unsigned long>(xbt_cfg_get_double("smpi/shared-malloc-blocksize"));
+  void *mem, *allocated_ptr;
+  size_t allocated_size;
+#ifdef MAP_HUGETLB
```
mquinson: You can factorize this if block by defining use_huge_page to false when not MAP_HUGETLB

src/smpi/smpi_shared.cpp
```diff
@@ -218,23 +223,63 @@ static void *smpi_shared_malloc_local(size_t size, const char *file, int line)
...
   /* First reserve memory area */
-  mem = mmap(NULL, size, PROT_READ | PROT_WRITE, MAP_ANONYMOUS | MAP_PRIVATE, -1, 0);
+  allocated_ptr = mmap(NULL, allocated_size, PROT_READ | PROT_WRITE, MAP_ANONYMOUS | MAP_PRIVATE, -1, 0);
 
-  xbt_assert(mem != MAP_FAILED, "Failed to allocate %zuMiB of memory. Run \"sysctl vm.overcommit_memory=1\" as root "
+
+  xbt_assert(allocated_ptr != MAP_FAILED, "Failed to allocate %zuMiB of memory. Run \"sysctl vm.overcommit_memory=1\" as root "
                                 "to allow big allocations.\n",
              size >> 20);
+#ifdef MAP_HUGETLB
```
mquinson: same factorization needed here

mquinson:
Also, you need to document the option (and how to use it) in doc/doxygen/options.doc, please.
Thanks,

src/smpi/smpi_shared.cpp
Ezibenroc: When huge pages are used, there are still some "classical" calls to mmap without `MAP_HUGETLB` for areas smaller than a huge page (the start and the end of the blocks).
Apparently, you cannot call mmap without `MAP_HUGETLB` on a file opened in a `hugetlbfs` mount point. So you need two files in this case.

Ezibenroc:
Good point for the factorization, it is more readable now.

I added the documentation for the huge page inside the section of shared malloc and not in a new section, because I think that it is more relevant here. I also added the documentation for the partial shared malloc that was missing.

src/smpi/smpi_shared.cpp
mquinson: Damn, Ok. I think you should add a little comment about it, maybe.

doc/doxygen/options.doc
```diff
@@ -1115,6 +1115,35 @@ only consume 1 MiB in memory.
...
+
+\code{.unparsed}
+    mkdir /home/huge
+    sudo mount none /home/huge -t hugetlbfs -o rw,mode=0777
+    sudo sh -c 'echo 1 > /proc/sys/vm/nr_hugepages'
+\endcode
+
+Then, you can pass the option --cfg=smpi/shared-malloc-hugepage:/home/huge
+to smpirun.
+
```
mquinson: Very good, thanks. You could improve it by saying that "echo 1" is the minimal size that we need but that any larger value is ok, but that's not mandatory. It's really good as is. Instructive and succint.

mquinson:
We still don't know how to test it, though. We create a hugetlbfs mountpoint to a specific location (like ~/hugepages) on several jenkins hosts, and have the cmake using it if it exists.

Or we try to have cmake using /dev/hugepages by default, which rights should be changed on the  jenkins hosts.

@degomme, what do you think?

Ezibenroc:
Correct me if I missed something, but this feature is still not tested. We should do something.

mquinson:
I agree we should test it. But how? We could change the permission on /dev/hugepages on some hosts on Jenkins, and have a test trying to use it if possible? This could do the trick. 

I'm reopening+retitling the issue until this is fixed.

Thanks for the poke.

PR #177, opened 2017-05-31
Use automated call to sd_exit

Takishipp:


mquinson:
Hello,

I don't see the onexit() call. Do I miss something?

Thx,

mquinson:
It would have been cool if you could have checked how it's done for MSG_exit() as hinted in the bug report #3.
https://github.com/simgrid/simgrid/blob/master/src/msg/msg_global.cpp#L69
 Please honorate the clean-atexit configuration option for SimDag also.

PR #179, opened 2017-06-02
call sg_instr_new_host via signal call

Takishipp:


src/surf/xml/platf_private.hpp
```diff
@@ -247,7 +247,7 @@ XBT_PUBLIC(void) routing_route_free(sg_platf_route_cbarg_t route);
 /********** Instr. **********/
 XBT_PRIVATE void sg_instr_AS_begin(sg_platf_AS_cbarg_t AS);
 XBT_PRIVATE void sg_instr_new_router(const char* name);
-XBT_PRIVATE void sg_instr_new_host(simgrid::s4u::Host& host);
+void sg_instr_new_host(simgrid::s4u::Host& host);
```
mquinson: Please remove that line. I think you dont need it anymore

PR #180, opened 2017-06-06
sg_instr_new_router with netpoint signal

Takishipp:


mquinson:
Please add a signal to the Router type, and get your function registered to that signal.

PR #181, opened 2017-06-07
Add tests for multi-core VM

bcamus:
I add some tests to check the behavior of the new multi-core VM feature. Please note that some of these tests currently fail : either I made some mistakes when defining the expected outputs, or the current behavior is wrong.


mquinson:
Hello, you are changing the file examples/platforms/cluster.xml, which is not a good idea. Could you please undo this change?

Also, it'd be cool if you could fix the warning raised by codacy (a variable initialized twice).

Thanks a lot.

bcamus:
Ok, I made the requested changes.

PR #202, opened 2017-07-18
Turn paje_value to a class

Takishipp:


src/instr/instr_interface.cpp
```diff
@@ -176,7 +176,7 @@ void TRACE_declare_mark(const char *mark_type)
  * \see TRACE_mark
  */
 void TRACE_declare_mark_value_with_color (const char *mark_type, const char *mark_value, const char *mark_color)
-{
+{ paje_value pj_value;
```
mquinson: Could you please follow the advices of http://simgrid.gforge.inria.fr/simgrid/latest/doc/uhood_tech.html (in particular the section about the coding standards and clang-format)?

src/instr/instr_private.h
```diff
@@ -262,6 +262,14 @@ class NewEvent : public PajeEvent  {
 
 };
 
+class paje_value{
```
mquinson: Seriously?? Is this code object oriented in any way? Where are the constructors, the instances and the destructor?

Please make val_t a class, change PJ_value_new into the constructor of that class, and the two other ones into static methods. And while you're at it, make every fields private, and add getter methods.

PR #203, opened 2017-07-19
Fix smpi_execute_benched.

Ezibenroc:
When simulating a target machine T on an host H, you had to give the time as measured on H. This time was then “converted” to the time that would be spent on T by using the host speed and the target speed.

Now, you directly give the time that should be spent on T. This is better, since the relative duration of different pieces of code may differ from H to T (and thus you really want to give the time spent on T).


mquinson:
shouldn't that be documented in some way? I'm not sure of whether the previous behavior was, actually :(

Ezibenroc:
I don't think this function was documented, but yes I guess it should be.

Where should I add it? In `doc/doxygen/options.doc`?
And maybe we should reconsider the name of the function while we are at it, it is not very explicit. Maybe `smpi_execute_duration`?

mquinson:
So this function is neither documented nor tested, right? Mwokay... Not your fault...

If the old name is bad for the new behavior, shouldn't we add your new function alongside to the old one?

Sorry for being picky, I try to understand...

Ezibenroc:
> So this function is neither documented nor tested, right?

Right. For the record, it was added with #167. 
I will write an integration test. I think a unit test would also be good, but I do not know where to put it, as `grep -Ri xbt_test src/smpi/` returns nothing (and if I recall well, I read somewhere that new unit tests should not be made with xbt_test anymore). Any idea?

> If the old name is bad for the new behavior, shouldn't we add your new function alongside to the old one?

Well, I do not think there is any use case for the old behavior. So this would simply be a dead part of the API, which is not wishable.

mquinson:
You are right, and right. 

A unit test would be good, but you don't want to use the old, deprecated, XBT mechanism. Have a look at `src/surf/trace_mgr_test.cpp` instead: it's using the boost unit testing library. It's compiled from the end of `tools/cmake/Tests.cmake`. I should take the time to put these information in `doc/doxygen/inside_tests.doc` at some point. Sorry for this sorry state.

And you're right, there is no point of keeping unused functions.

mquinson:
So, @Ezibenroc you want me to merge this, or you want to add the tests first?

Thanks,

Ezibenroc:
No I would like to add tests first, I will try tomorrow.

Ezibenroc:
We need the fixed `smpi_execute_benched` to be in the master branch for a paper, but I did not take the time to write tests.
So I am merging this and opening an issue.

Note that the CI errors are due to timeouts in tests not impacted by this PR.

PR #208, opened 2017-07-26
Delete some useless lines

Takishipp:


src/instr/instr_smpi.h
```diff
@@ -29,7 +29,6 @@ XBT_PRIVATE void TRACE_smpi_testing_out(int rank);
 XBT_PRIVATE void TRACE_smpi_testing_in(int rank, instr_extra_data extra);
 XBT_PRIVATE void TRACE_smpi_alloc();
 XBT_PRIVATE void TRACE_smpi_release();
-XBT_PRIVATE void TRACE_smpi_ptp_in(int rank, const char* operation, instr_extra_data extra);
```
mquinson: Unfortunately, you cannot merge these events. ptp means Point to Point, and that's different from collective. This is true that this is the same behavior for both of them, but actually, each of them should be changed into a signal, at some point.

src/instr/instr_paje_types.cpp
```diff
@@ -52,7 +52,7 @@ void PJ_type_free (type_t type)
   xbt_dict_cursor_t cursor = nullptr;
   xbt_dict_foreach (type->values, cursor, value_name, val) {
     XBT_DEBUG("free value %s, child of %s", val->name, val->father->name);
-    delete val;
+    free(val);
```
mquinson: Good catch, the value is not created by new and thus should not be freed with delete. I think that the best solution is to actually use new and delete, but this is already an improvement, thanks.

src/instr/instr_paje_values.cpp
```diff
@@ -31,9 +31,9 @@ value::value(const char* name, const char* color, type_t father)
 
 value::~value()
 {
-  xbt_free(name);
+/*   xbt_free(name);
```
mquinson: This should not be commented. These variables were created in the constructor and should thus be freed in the destructor.

mquinson:
I integrated some parts manually, and the rest should be done another way. Thus closing, sorry.

PR #212, opened 2017-08-02
Value cleanup

Takishipp:


src/instr/instr_paje_values.cpp
```diff
@@ -15,27 +15,21 @@ value::value(const char* name, const char* color, type_t father)
...
-  XBT_DEBUG("new value %s, child of %s", ret->name, ret->father->name);
-  LogEntityValue(this->ret);
+  xbt_dict_set (father->values, name, this, nullptr);
+  XBT_DEBUG("new value %s, child of %s", name, father->name);
+  LogEntityValue(this);
 };
 
 value::~value()
 {
-  /* FIXME: this should be cleanable
```
mquinson: Please don't remove this but uncomment it

src/instr/instr_private.h
```diff
@@ -72,13 +72,8 @@ class s_type {
 typedef s_type s_type_t;
 
 //--------------------------------------------------
-class value {
+class value : public s_type {
```
mquinson: Please don't make value a children of type (yet)

src/instr/instr_paje_types.cpp
```diff
@@ -52,7 +52,6 @@ void PJ_type_free (type_t type)
   xbt_dict_cursor_t cursor = nullptr;
   xbt_dict_foreach (type->values, cursor, value_name, val) {
     XBT_DEBUG("free value %s, child of %s", val->name, val->father->name);
-    xbt_free(val);
```
mquinson: You need to delete val, not to remove the line!

src/instr/instr_config.cpp
```diff
@@ -150,7 +150,6 @@ int TRACE_end()
 
     /* destroy all data structures of tracing (and free) */
     PJ_container_free_all();
-    recursiveDestroyType (PJ_type_get_root());
```
mquinson: If you don't do this, how will your data get freed? Please restore this line.

src/instr/instr_paje_types.cpp
```diff
@@ -45,33 +45,14 @@ static type_t newType (const char *typeNameBuff, const char *key, const char *co
   return ret;
 }
 
-void PJ_type_free (type_t type)
+s_type::~s_type()
 {
-  value* val;
-  char *value_name;
-  xbt_dict_cursor_t cursor = nullptr;
-  xbt_dict_foreach (type->values, cursor, value_name, val) {
```
mquinson: How will the values get freed if you don't free them? You should use delete instead of xbt_free, but this really must be done.

src/instr/instr_paje_types.cpp
```diff
@@ -45,33 +45,14 @@ static type_t newType (const char *typeNameBuff, const char *key, const char *co
...
-  xbt_dict_free (&type->values);
-  xbt_free (type->name);
-  xbt_free (type->id);
-  xbt_free (type->color);
-  xbt_dict_free (&type->children);
-  xbt_free (type);
-  type = nullptr;
-}
-
-void recursiveDestroyType (type_t type)
```
mquinson: The types are stored in a recursive data structure. If you remove the function in charge of recursively destroy these objects, how do you think they will get freed?

src/instr/instr_private.h
```diff
@@ -72,10 +72,10 @@ class s_type :public ess_type {
   e_entity_types kind;
   xbt_dict_t children;
   xbt_dict_t values; //valid for all types except variable and container
+  s_type(const char *typeNameBuff, const char *key, const char *color, e_entity_types kind, type_t father);
```
mquinson: I'm happy you're working to cleanup the s_type but it should not be part of the same PR.

src/instr/instr_paje_types.cpp
Takishipp: But according to the class, the son cannot be exist without the existence of the father, am I wrong?

src/instr/instr_paje_types.cpp
Takishipp: How can I use delete instead of xbt_free, while the allocation is done with calloc?

src/instr/instr_paje_types.cpp
mquinson: The son will not be automatically be removed if you remove the father. You have to do it explicitely.

src/instr/instr_paje_types.cpp
mquinson: The best approach is to use "new" instead of calloc.

mquinson: Just to be clear: I would like the content of this->values to be cleaned up. For that, you have to restore the ```xbt_dict_foreach (type->values, cursor, value_name, val) {``` part, and use ```delete value``` in it.

mquinson:
Could you please fix the msg-trace-masterworker test that is segfaulting? Thanks in advance.

mquinson:
This was done another way in between. Thanks anyway.

PR #214, opened 2017-08-07
Simgrid for MPI load balancing simulation

rktesser:
@mquinson Here is my pull request #125 squashed into a single commit.

### DIFF
--- a/src/instr/instr_paje_trace.cpp
+++ b/src/instr/instr_paje_trace.cpp
@@ -27,6 +27,8 @@ FILE *tracing_file = nullptr;
 static xbt_dict_t tracing_files = nullptr; // TI specific
 static double prefix=0.0; // TI specific
 
+static xbt_dynar_t state_tracker = nullptr; //TI specific
+
 std::vector<PajeEvent*> buffer;
 void buffer_debug(std::vector<PajeEvent*> *buf);
 
@@ -571,11 +573,12 @@ void PushStateEvent::print() {
     instr_extra_data extra = (instr_extra_data)extra_;
 
     char* process_id = nullptr;
-    // FIXME: dirty extract "rank-" from the name, as we want the bare process id here
-    if (strstr(container->name, "rank-") == nullptr)
+    // FIXME: dirty extract "rank-" or "hostname-rank-" from the name, as we want the bare process id here
+    char *dash_pos = strrchr(container->name, '-');
+    if (dash_pos == nullptr)
       process_id = xbt_strdup(container->name);
     else
-      process_id = xbt_strdup(container->name + 5);
+      process_id = xbt_strdup(dash_pos + 1);
 
     FILE* trace_file = (FILE*)xbt_dict_get(tracing_files, container->name);
 
@@ -587,22 +590,22 @@ void PushStateEvent::print() {
         fprintf(trace_file, "%s finalize\n", process_id);
         break;
       case TRACING_SEND:
-        fprintf(trace_file, "%s send %d %d %s\n", process_id, extra->dst, extra->send_size, extra->datatype1);
+        fprintf(trace_file, "%s send %d %d %d %s\n", process_id, extra->dst, extra->tag, extra->send_size, extra->datatype1);
         break;
       case TRACING_ISEND:
-        fprintf(trace_file, "%s Isend %d %d %s\n", process_id, extra->dst, extra->send_size, extra->datatype1);
+        fprintf(trace_file, "%s Isend %d %d %d %s\n", process_id, extra->dst, extra->tag, extra->send_size, extra->datatype1);
         break;
       case TRACING_RECV:
-        fprintf(trace_file, "%s recv %d %d %s\n", process_id, extra->src, extra->send_size, extra->datatype1);
+        fprintf(trace_file, "%s recv %d %d %d %s\n", process_id, extra->src, extra->tag, extra->send_size, extra->datatype1);
         break;
       case TRACING_IRECV:
-        fprintf(trace_file, "%s Irecv %d %d %s\n", process_id, extra->src, extra->send_size, extra->datatype1);
+        fprintf(trace_file, "%s Irecv %d %d %d %s\n", process_id, extra->src, extra->tag, extra->send_size, extra->datatype1);
         break;
       case TRACING_TEST:
         fprintf(trace_file, "%s test\n", process_id);
         break;
       case TRACING_WAIT:
-        fprintf(trace_file, "%s wait\n", process_id);
+        fprintf(trace_file, "%s wait %d %d %d\n", process_id, extra->src, extra->dst, extra->tag);
         break;
       case TRACING_WAITALL:
         fprintf(trace_file, "%s waitAll\n", process_id);
@@ -670,6 +673,10 @@ void PushStateEvent::print() {
       case TRACING_ALLGATHER: // rank allgather sendcount recvcounts (sendtype) (recvtype)
         fprintf(trace_file, "%s allGather %d %d %s %s", process_id, extra->send_size, extra->recv_size, extra->datatype1, extra->datatype2);
         break;
+      case TRACING_CUSTOM:
+	if(extra->print_push)
+	  extra->print_push(trace_file, process_id, extra);
+	break;
       case TRACING_WAITANY:
       case TRACING_SENDRECV:
       case TRACING_SCATTER:
@@ -685,13 +692,15 @@ void PushStateEvent::print() {
         XBT_WARN("Call from %s impossible to translate into replay command : Not implemented (yet)", val->name);
         break;
     }
+ 
+    //We need to keep track of the current state, so we can register when the
+    //current iteration (TRACING_ITERATION) ends. 
+    // We will push the whole 'extra', because we need access to the printing
+    // function when the state is popped.
+    xbt_dynar_push(state_tracker, &extra);
 
-    if (extra->recvcounts != nullptr)
-      xbt_free(extra->recvcounts);
-    if (extra->sendcounts != nullptr)
-      xbt_free(extra->sendcounts);
     xbt_free(process_id);
-    xbt_free(extra);
+    //'extra' will be freed in PopStateEvent::print
 
   } else {
     THROW_IMPOSSIBLE;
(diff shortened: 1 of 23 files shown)


mquinson:
Many thanks for the hard work. I will review it tomorrow. Could you please rebase it on top of the current git trunk? Thanks in advance.

rktesser:
Rebase done!

rktesser:
I see some tests are failing on Travis CI, but I can't reproduce the errors in my computer. For me all tests are passed.

mquinson:
Only the smpi-tracing-call-location test is relevant to your PR. The other one can be ignored.
You should try running it in valgrind, but this test is a bit difficult to valgrind, because it's in several parts: one command line produces a trace, and the second one reads the trace. The failing test is often the second one, but you have to produce the trace first. It is not enough to run ```ctest -R smpi-tracing-call-location``` as the trace file is removed by the end of the command.

Another solution is to have a build directory (see http://simgrid.gforge.inria.fr/simgrid/latest/doc/install.html#install_cmake_outsrc and http://simgrid.gforge.inria.fr/simgrid/latest/doc/uhood_tech.html) with the build option ```-Denable_memcheck=ON``` This will run valgrind on each and every of our tests, breaking most of them. Any test having a memleak will be reported as failing, but at least, you'll get the info you need on ```smpi-tracing-call-location```

rktesser:
This test does not fail in the two machines I tested. So, I can tell you for sure that the generated trace is correct. The problem is probably with the test itself failing.

I will try to run with memory check. I already have a separate build directory. I can create another one for this. Or I may simply run the commands used in the '.tesh' file manually (that's what I did to debug the tests that were crashing before).

rktesser:
From my tests, it seems that using valgrind changes the order in which the traces are written. In the case of   the smpi-tracing-call-location, the test assumes that the identifier for PMPI_Barrier will not change (it simply does a grep on the trace). But when running with valgrind this event type is registered later than without valgrind, causing its identifier to change. This can be seen by looking at these two traces (one without and one with valgrind:
[smpi_trace-no_valgrind.txt](https://github.com/simgrid/simgrid/files/1212830/smpi_trace-no_valgrind.txt)
[smpi_trace-with_valgrind.txt](https://github.com/simgrid/simgrid/files/1212833/smpi_trace-with_valgrind.txt).

I will change the test to actually get the identifier from the trace. Then, let's se if this test still fails.

rktesser:
I tried to modify the tests to be more dynamic, but it seems tesh does not support stream redirection ('|', '<','>'...), which makes this difficult. The problem is that I need to run a command to get te id of the PMPI_Barrier event and then use it in a second command, to check if the call location is correct.

I could, maybe, make tesh generate an external script (e.g. a Python script) to do the test. Would this be acceptable?

mquinson:
Test again. I think that @frs69wq fixed the changing order of events. If it's not enough, we should find the unordered_map that needs to be changed into a map.
I don't think that fixing the test is the easiest approach here.

rktesser:
Hi. I jus did a new rebase. The problem still exists.

src/simgrid/host.cpp
```diff
@@ -106,6 +106,24 @@ xbt_dynar_t sg_hosts_as_dynar()
   return res;
 }
 
+xbt_dynar_t sg_host_get_processes_as_dynar(sg_host_t host)
```
Shurakai: This can be replaced by Host::actorList(vector)

src/simgrid/host.cpp
```diff
@@ -106,6 +106,24 @@ xbt_dynar_t sg_hosts_as_dynar()
...
+  xbt_swag_t process_swag = host->extension<simgrid::simix::Host>()->process_list;
+  
+  xbt_swag_foreach(process, process_swag){
+    xbt_dynar_push(process_dynar, &process);
+  }
+  
+  return process_dynar;
+}
+
+int sg_host_get_process_count(sg_host_t host){
```
Shurakai: We should probably introduce a method in S4U that does this.

src/instr/instr_private.h
```diff
@@ -398,7 +398,8 @@ typedef enum{
   TRACING_COMPUTING,
   TRACING_SLEEPING,
   TRACING_SCAN,
-  TRACING_EXSCAN
+  TRACING_EXSCAN,
+  TRACING_CUSTOM
 } e_caller_type ;
 
 typedef struct s_instr_extra_data {
```
Shurakai: This whole file and this data type doesn't exist anymore; I think this is now in instr_private.hpp organized as a class: TIData. (With renamed attributes, e.g. datatype1 -> send_type, ...)

src/instr/instr_smpi.h
```diff
@@ -35,7 +35,10 @@ XBT_PRIVATE void TRACE_smpi_send(int rank, int src, int dst, int tag, int size);
 XBT_PRIVATE void TRACE_smpi_recv(int src, int dst, int tag);
 XBT_PRIVATE void TRACE_smpi_init(int rank);
 XBT_PRIVATE void TRACE_smpi_finalize(int rank);
-XBT_PRIVATE char* smpi_container(int rank, char* container, int n);
+XBT_PRIVATE void TRACE_smpi_send_process_data_in(int rank);
+XBT_PRIVATE void TRACE_smpi_send_process_data_out(int rank);
+XBT_PRIVATE void TRACE_smpi_process_change_host(int, sg_host_t, sg_host_t, int);
+XBT_PUBLIC(char*) smpi_container(int rank, char* container, int n);
```
Shurakai: This is an important change that I need to verify; I'd rather not make this public.

src/smpi/bindings/smpi_pmpi_request.cpp
```diff
@@ -91,8 +145,14 @@ int PMPI_Start(MPI_Request * request)
...
     retval = MPI_ERR_REQUEST;
   } else {
-    (*request)->start();
-    retval = MPI_SUCCESS;
+    if((*request)->flags() & SEND){
+      retval = persistent_send(request);
+    }else if((*request)->flags() & RECV){
+      retval = persistent_recv(request);
+    }else{
+      (*request)->start();
```
Shurakai: I'm not sure if we need the else clause here. The standard says:

> A communication (send or receive) that uses  a  persistent  request  is initiated by the function MPI_Start.

Also check whether the request is a persistent one (see standard).

src/smpi/include/private.h
```diff
@@ -60,7 +60,7 @@ typedef SMPI_Graph_topology *MPIR_Graph_Topology;
 
 typedef SMPI_Dist_Graph_topology *MPIR_Dist_Graph_Topology;
 
-XBT_PRIVATE SMPI_Process* smpi_process();
+XBT_PUBLIC(SMPI_Process*) smpi_process();
```
Shurakai: Make sure these private to public changes are really necessary; it's pretty ugly.

src/smpi/internals/smpi_replay.cpp
```diff
@@ -31,22 +37,74 @@ char* sendbuffer=nullptr;
...
-  }
+
+/******************************************************************************
+ * This code manages the request dictionaries which are used by the
+ * assynchronous comunication actions (Isend, Irecv, wait, test).
+ *****************************************************************************/
+
+static std::string build_request_key(int src, int dst, int tag){
+  std::stringstream key_stream;
+  key_stream << src << dst << tag;
```
Shurakai: Here's another problem: If we really need to generate this key, we need to distinguish between messages. Such as:

1111 (src = 1, dst = 11, tag = 1) and the message
1111 (src = 1, dst = 1, tag = 11)

src/smpi/internals/smpi_replay.cpp
```diff
@@ -373,24 +464,29 @@ static void action_test(const char *const *action){
...
 
 static void action_wait(const char *const *action){
-  CHECK_ACTION_PARAMS(action, 0, 0)
+  CHECK_ACTION_PARAMS(action, 3, 0);
+  int src = atoi(action[2]);
+  int dst = atoi(action[3]);
+  int tag = atoi(action[4]);
   double clock = smpi_process()->simulated_elapsed();
+  std::string key = build_request_key(src, dst, tag);
+  MPI_Request request = get_request_with_key(key);
```
Shurakai: This is probably the reason why the request is stored with a string as the key: In this case, we don't seem to have any request and so it's impossible to easily retrieve the request? However, the old version just accessed the last element (see the diff). Why is this not true any longer?

src/smpi/internals/smpi_replay.cpp
```diff
@@ -172,8 +235,12 @@ static void action_init(const char *const *action)
   smpi_process()->simulated_start();
   /*initialize the number of active processes */
   active_processes = smpi_process_count();
-
-  set_reqq_self(new std::vector<MPI_Request>);
+  if(reqd.empty()){
+    for(int i = 0; i< active_processes; i++){ 
+      std::unordered_map<std::string, MPI_Request> req_map;
+      reqd.push_back(req_map);
```
Shurakai: This will insert a copy, meaning req_map can be moved out of the loop.

src/instr/instr_paje_trace.cpp
```diff
@@ -587,22 +590,22 @@ void PushStateEvent::print() {
...
       case TRACING_IRECV:
-        fprintf(trace_file, "%s Irecv %d %d %s\n", process_id, extra->src, extra->send_size, extra->datatype1);
+        fprintf(trace_file, "%s Irecv %d %d %d %s\n", process_id, extra->src, extra->tag, extra->send_size, extra->datatype1);
         break;
       case TRACING_TEST:
         fprintf(trace_file, "%s test\n", process_id);
         break;
       case TRACING_WAIT:
-        fprintf(trace_file, "%s wait\n", process_id);
+        fprintf(trace_file, "%s wait %d %d %d\n", process_id, extra->src, extra->dst, extra->tag);
```
Shurakai: Printing the tags has been integrated in instr_private.hpp / Pt2PtTIData

src/instr/instr_paje_trace.cpp
```diff
@@ -720,7 +729,26 @@ void PopStateEvent::print() {
...
-    /* Nothing to do */
+    char *process_id;
+
+    instr_extra_data extra;
+    xbt_dynar_pop(state_tracker, &extra);
+    if(extra->type  == TRACING_CUSTOM){
+      if (strstr(container->name, "rank-") == nullptr)
+	process_id = xbt_strdup(container->name);
+      else
+	process_id = xbt_strdup(container->name + 5);
```
Shurakai: This if/else should simply be this_actor::get_pid();

src/instr/instr_paje_trace.cpp
```diff
@@ -720,7 +729,26 @@ void PopStateEvent::print() {
...
+
+      FILE *trace_file = (FILE *)xbt_dict_get(tracing_files, container->name);
+      if(extra->print_pop){
+        extra->print_pop(trace_file, process_id, extra);
```
Shurakai: We should not have to care about this anymore; all we need is a custom subclass of TIData that is used instead of the "extra" structure; in this case, for instance in PMPI_Iteration_in() etc. (not implemented here, implemented in the LB code of rktesser).

src/simgrid/host.cpp
Shurakai: Host::get_actor_count has been implemented.

src/smpi/bindings/smpi_pmpi_request.cpp
```diff
@@ -83,6 +83,60 @@ int PMPI_Ssend_init(void* buf, int count, MPI_Datatype datatype, int dst, int ta
...
+  TRACE_smpi_ptp_in(rank, __FUNCTION__, extra);
+
+  (*request)->start();
+  retval = MPI_SUCCESS;
+
+  TRACE_smpi_ptp_out(rank, rank, __FUNCTION__);
+
+  return retval;
+}
+
```
Shurakai: This was all incorporated in 26713cb93739af3d63913a251aeeccf433f8da7a

Shurakai:
Basically all of this has been integrated and pushed to the master branch in numerous commits.

Thanks @rktesser . You can close this, @mquinson

mquinson:
Many thanks to @rktesser for the initial implementation, and also a huge thanks to @Shurakai for the very hard job of integrating this work into the codebase !!

PR #216, opened 2017-08-17
s_container cleanup

Takishipp:


frs69wq:
Hi,
Sorry to put my nose in your work, but for the last few months we strive to make SimGrid code more readable and written in proper C++. What I see in this PR is way too far from the coding standards we try to apply to the whole code: Naming conventions of classes are not respected (please look to existing classes for example), functions are merely renamed, but this doesn't make them class methods, Object Oriented Programming principles are not respected at all ... 
A glaring example is this:  +void s_container::s_container_remove_from_parent (container_t child) 
* child should be caller of a removeFromParent() method of the container class
* s_container is the old fashioned C naming of  structs. Now, in C++ we prefer Container as class name 
* container_t is the old fashioned C naming of pointers on structs. Now, in C++ we prefer Container* 
* This container class has no destructor (while there is allocated memory with the dict)
* This container class has a children field, but also a type_t type filed, that also contains a dictionnary of children. If this correspond to the same information, one has to be removed as soon as possible to prevent later issue when deallocating memory.
* By the way, const char* should be replaced by C++ std::string and xbt_dict_t by std::unordered_map, but that is another story.

Thanks for trying to follow the way the code is written is the rest of SimGrid. I'm not speaking of other files in directories you never looked at, but at least of what is in src/instr/instr_private.h (which should become a hpp file at some point by the way).

Takishipp:
Hi, 

I would like to thank you for the useful remarks that I will take in count in my further PR. 

Best regards. 

----- Mail original -----

mquinson:
this is superseeded by the code I wrote in between, thus closing. Thanks for the fish.

PR #217, opened 2017-08-18
Cleanup

Takishipp:


mquinson:
This was done another way in between. Sorry.

PR #224, opened 2017-09-22
Add s4u-async-wait to the examples of s4u 

Takishipp:
The example doesn't works well, I tried to figure out but I'm always getting a deadlock.


mquinson:
Hello, 

you need to fix the issues with this code, please. In particular, rebase your work on top of the current trunk, and make sure that codacy does not complain anymore (also travis checks need to agree on your code).

Thanks, Mt.

mquinson:
Thanks for this!

PR #226, opened 2017-10-05
Convert process-create to s4u is complete

Takishipp:


mquinson:
This is a dupplication of the existing actor-create/s4u-actor-create.cpp example, so I will not integrate it. Sorry.

Takishipp:
I think that they aren't the same because actor-create give users how to make a simple exchange between sender and receiver, and process-create show how to deal with a task that could be eventually be sent or received.

mquinson:
Well, no. the example name is about process-create. The documentation is about process create, and the code is about executing stuff once the process is started. So, I fail to see the connection to task exchange...

Moreover, processes are called "actors" in S4U, and there is no such thing like task in S4U. Check the app-pingpong/s4u-app-pingpong.cpp example to see how data is exchanged in S4U.

PR #236, opened 2017-11-08
Use unsigned int for std:*map positions (must be >0).

mazhe:
degomme approves this commit :)


mquinson:
if degomme likes it, then I'll merge it :)

Thanks,

Issue #223, opened 2017-09-08
MSG_task_get_flops_amount() is not documented


mickours:
Hi,

I am currently implementing job preemption in Batsim (that is based on Simgrid). To reflect the preemption cost properly I'd like to have access to the remaining (or already computed) amount of work of an MSG ptasks before killing it.

I found a ``getRemains()`` function used in ``surf/ptask_L07.cpp`` that seems to correspond to what I need but it is not accessible in MSG. Is it possible to expose this in the MSG API?

mickours:
I finally find a way to get what I need from MSG using ``MSG_task_get_flops_amount(task);`` that returns exactly what what I was looking for when the task is started: the remaining amount of work between 0 and 1 for a parallel task.
I did not find it at first, because It not documented.

mquinson:
This function should actually be documented...

mickours:
The problem is that this function actually returns the flops amount (but not the parallel flops amount) before the task start, but when the task has started it return the remaining amount of work of the task (for parallel also).

So we should change the behavior of this function to return only the flops amount and create another function that return the remaining works directly. But it is not *only* documentation and it breaks the existing API.

mquinson:
If I understand correctly, you propose to:
 - [  ] change  `MSG_task_get_flops_amount()` so that it always return the initial amount 
 - [  ] add a `MSG_task_get_flops_remaining()` with the current behavior of  `MSG_task_get_flops_amount()`

What's not clear to me is whether there is something to do for the parallel tasks specifically.

frs69wq:
The problem with parallel tasks is that it's up to the user to define 
how the original amount given on task creation is converted into the 
combo "array of flops + matrix of bytes". For instance in SimDag, it's 
possible to use Amdahl's Law for that, but it can be anything. Then, I 
don't see how SimGrid could return something meaningful for the 
evolution of parallel tasks.


Le 01/10/2017 à 01:38, Martin Quinson a écrit :

mickours:
Has I said, after the task begins, It do not return the amount of flops at all: It returns the remaining amount of work which is a value between 0 and 1. This value comes directly from the underlying model and works for either parallel or sequential tasks and for communication and computation (even it comes from the compute object: this is the only used for parallel task).

I put the code here for convenience:
https://github.com/simgrid/simgrid/blob/934e15f8ffc473c09214dd3dcbd4a926d2fa0846/src/msg/msg_task.cpp#L231-L237
 
[...]

PR #237, opened 2017-11-09
 [MSG] add MSG_task_get_remaining_work_ratio + test

mickours:
See https://github.com/simgrid/simgrid/issues/223 for details.

WARNING: It breaks the API by changing the behavior of MSG_task_get_flops_amount (to a more sane one...)

### DIFF
--- a/examples/msg/cloud-capping/cloud-capping.c
+++ b/examples/msg/cloud-capping/cloud-capping.c
@@ -76,8 +76,8 @@ static void test_dynamic_change(void)
   MSG_process_create("worker0", worker_busy_loop_main, &task0, (msg_host_t)vm0);
   MSG_process_create("worker1", worker_busy_loop_main, &task1, (msg_host_t)vm1);
 
-  double task0_remain_prev = MSG_task_get_flops_amount(task0);
-  double task1_remain_prev = MSG_task_get_flops_amount(task1);
+  double task0_remain_prev = MSG_task_get_remaining_work_ratio(task0);
+  double task1_remain_prev = MSG_task_get_remaining_work_ratio(task1);
 
   const double cpu_speed = MSG_host_get_speed(pm0);
   for (int i = 0; i < 10; i++) {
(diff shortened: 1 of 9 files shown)


mquinson:
Hello, thanks for that change. It definitively goes in the right direction. You simply should compile with enable_compile_warnings enabled :) http://simgrid.gforge.inria.fr/simgrid/latest/doc/uhood_tech.html

Thanks to your work, I now see how to sort out the mess without breaking the API. Could you please:
- restore MSG_task_get_flops_amount() to its borken implementation (to not break the code of old users)
- Remove the doc of that function (to hide it from new users)
- Add a XBT_ATTRIB_DEPRECATED_v321() marker to that function (to hint current users are prepare future cleanups)
- Add a MSG_task_get_flops() function with the right implem (the one you propose in your patch), along with the doc and everything.
- You also have some minor details pointed by codacy.

That would be really cool because it would move in the right direction, without being too intrusive to our users.

Thanks in advance, 
Mt.

PS: if it bothers you too much, I could probably do it myself, I have a long train trip planned soon. Just drop me an email.

mickours:
I changed the code to reflect your propositions.

To sum up:
1. [X] restore MSG_task_get_flops_amount() to its borken implementation
2. [X] Remove the doc of that function
3. [ ] Add a XBT_ATTRIB_DEPRECATED_v321() marker to that function: I was unable to compile that without -Werror=deprecated-declarations, so I leave it a comment in the code if you find a way. 
4. [X] Add a MSG_task_get_flops() function with the right implem:  I choose to add two functions instead:
  - ``MSG_task_get_initial_flops_amount`` with initial in it, to be clear that this value is static
  -  ``MSG_task_get_remaining_work_ratio`` that return the remaining amount of work between 1 and 0
5. [X] Remove warnings
6. [X] Patch other parts of Simgrid code to reflect this API change

mquinson:
Many thanks for that, that's really a nice change!

PR #238, opened 2017-11-20
remove ununsed exception in javadocs

barais:
ununsed exception in javadocs create errors when building simgrid with last jdk8


mquinson:
Thanks! The failed test on travis is confirmed to be a false positive (that's a timeout on the MPI+rma tests, no way it's caused by your change on Java).

PR #242, opened 2017-11-22
S4U version of energy pstate example 

Takishipp:


examples/s4u/README.doc
```diff
@@ -101,6 +101,12 @@ documentation, but it should remain readable directly.
     @ref examples/s4u/actor-migration/s4u-actor-migration.cpp \n
     Actors can move or be moved from a host to another with the @ref migrate method.
 
+  - <b>Using Pstates on a host</b>
+    @ref examples/s4u/energy-pstate/s4u-energy-pstate.c\n
```
mquinson: This wont work if you don't add that file as an \@example at the bottom of the file.

PR #246, opened 2017-11-23
Fixed energy plugin: returning watts_off when host is off

danilo-carastan-santos:
fixed energy plugin: getCurrentWattsValue was not returning watts_of when the host is off


mquinson:
Thanks for that, good catch.

PR #256, opened 2018-03-12
Modernize FindSimGrid

Flamefire:
Enhancements:

- Use FindPackageHandleStandardArgs to properly accept find_package arguments
- Extract the version from  `SIMGRID_(MAJOR, MINOR, PATCH)`
- Define a target `SimGrid::SimGrid` in addition to the variables (modern CMake, variables should be deprecated...)
- Handle C++11 requirement as much as possible (note: If you stop using -std=gnu++11 you might need to set the property `CXX_EXTENSIONS OFF`: https://cliutils.gitlab.io/modern-cmake/chapters/cpp11.html)

### DIFF
--- a/FindSimGrid.cmake
+++ b/FindSimGrid.cmake
@@ -1,60 +1,92 @@
-# Cmake macro to search for the SimGrid library. 
+# CMake find module to search for the SimGrid library. 
 
-# Copyright (c) 2016. The SimGrid Team (version 20161210).
+# Copyright (c) 2016-2018. The SimGrid Team.
 #
 # This file is free software; you can redistribute it and/or modify it
 # under the terms of the license (GNU LGPL) which comes with this package.
 
-# The user can hint a path using the SimGrid_PATH option.
-#
 # Once done, the following will be defined:
-#  
-#  SimGrid_INCLUDE_DIR - the SimGrid include directories
-#  SimGrid_LIBRARY - link your simulator against it to use SimGrid
 #
-# Afterward, you should probably do the following:
+#  CMake >= 2.8.12:
+#    Target SimGrid::Simgrid
 #
-#  include_directories("${SimGrid_INCLUDE_DIR}" SYSTEM)
-#  target_link_libraries(your-simulator ${SimGrid_LIBRARY})
+#    Use as:
+#      target_link_libraries(your-simulator SimGrid::SimGrid)
 #
-# This could be improved in many ways (patches welcome):
-#  - Not finding SimGrid is fatal. But who would want to 
-#    overcome such a desperate situation anyway? ;)
-#  - No way to specify a minimal version
-#  - No proper find_package() integration
+#  Older CMake (< 2.8.12)
+#    SimGrid_INCLUDE_DIR - the SimGrid include directories
+#    SimGrid_LIBRARY - link your simulator against it to use SimGrid
+#
+#    Use as:
+#      include_directories("${SimGrid_INCLUDE_DIR}" SYSTEM)
+#      target_link_libraries(your-simulator ${SimGrid_LIBRARY})
+#
+# This could be improved:
+#  - Use automatic SimGridConfig.cmake creation via export/install(EXPORT in main CMakeLists.txt:
+#    https://cliutils.gitlab.io/modern-cmake/chapters/exporting.html
+#    https://cmake.org/Wiki/CMake/Tutorials/How_to_create_a_ProjectConfig.cmake_file
+#    https://github.com/boostcon/cppnow_presentations_2017/blob/master/05-19-2017_friday/effective_cmake__daniel_pfeifer__cppnow_05-19-2017.pdf
 
-set(SimGrid_PATH ${SimGrid_PATH} CACHE PATH "Path to SimGrid library and include")
+cmake_minimum_required(VERSION 2.8)
+
 find_path(SimGrid_INCLUDE_DIR
   NAMES simgrid_config.h
-  HINTS ${SimGrid_PATH}/include /usr/include /usr/local/include /opt/simgrid/include
+  PATHS /opt/simgrid/include
 )
 find_library(SimGrid_LIBRARY
   NAMES simgrid
-  HINTS ${SimGrid_PATH}/lib /usr/lib /usr/local/lib /opt/simgrid/lib
+  PATHS /opt/simgrid/lib
 )
 mark_as_advanced(SimGrid_INCLUDE_DIR)
 mark_as_advanced(SimGrid_LIBRARY)
 
-if (NOT EXISTS "${SimGrid_INCLUDE_DIR}")
-  message(FATAL_ERROR "Unable to find SimGrid header file. Please point the SimGrid_PATH variable to the installation directory.")
+if (SimGrid_INCLUDE_DIR)
+  set(SimGrid_VERSION_REGEX "^#define SIMGRID_VERSION_(MAJOR|MINOR|PATCH) ([0-9]+)$")
+  file(STRINGS "${SimGrid_INCLUDE_DIR}/simgrid_config.h" SimGrid_VERSION_STRING REGEX ${SimGrid_VERSION_REGEX})
+  set(SimGrid_VERSION "")
+
+  # Concat the matches to MAJOR.MINOR.PATCH assuming they appear in this order
+  foreach(match ${SimGrid_VERSION_STRING})
+    if(SimGrid_VERSION)
+      set(SimGrid_VERSION "${SimGrid_VERSION}.")
+    endif()
+    string(REGEX REPLACE ${SimGrid_VERSION_REGEX} "${SimGrid_VERSION}\\2" SimGrid_VERSION ${match})
+    set(SimGrid_VERSION_${CMAKE_MATCH_1} ${CMAKE_MATCH_2})
+  endforeach()
+  unset(SimGrid_VERSION_STRING)
+  unset(SimGrid_VERSION_REGEX)  
 endif ()
-if (NOT EXISTS "${SimGrid_LIBRARY}")
-  message(FATAL_ERROR "Unable to find SimGrid library. Please point the SimGrid_PATH variable to the installation directory.")
+
+include(FindPackageHandleStandardArgs)
+find_package_handle_standard_args(SimGrid
+  FOUND_VAR SimGrid_FOUND
+  REQUIRED_VARS SimGrid_INCLUDE_DIR SimGrid_LIBRARY
+  VERSION_VAR SimGrid_VERSION
+)
+
+if (SimGrid_FOUND AND NOT CMAKE_VERSION VERSION_LESS 2.8.12)
+  add_library(SimGrid::SimGrid SHARED IMPORTED)
+  set_target_properties(SimGrid::SimGrid PROPERTIES
+    INTERFACE_SYSTEM_INCLUDE_DIRECTORIES ${SimGrid_INCLUDE_DIR}
+    INTERFACE_COMPILE_FEATURES cxx_alias_templates
+    IMPORTED_LOCATION ${SimGrid_LIBRARY}
+  )
+  # We need C++11, so check for it
+  if (NOT CMAKE_VERSION VERSION_LESS 3.8)
+    # 3.8+ allows us to simply require C++11 (or higher)
+    set_property(TARGET SimGrid::SimGrid PROPERTY INTERFACE_COMPILE_FEATURES cxx_std_11)
+  elseif (NOT CMAKE_VERSION VERSION_LESS 3.1)
+    # 3.1+ is similar but for certain features. We pick just one
+    set_property(TARGET SimGrid::SimGrid PROPERTY INTERFACE_COMPILE_FEATURES cxx_alias_templates)
+  else ()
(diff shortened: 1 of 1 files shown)


FindSimGrid.cmake
```diff
@@ -1,60 +1,92 @@
...
+#    https://cmake.org/Wiki/CMake/Tutorials/How_to_create_a_ProjectConfig.cmake_file
+#    https://github.com/boostcon/cppnow_presentations_2017/blob/master/05-19-2017_friday/effective_cmake__daniel_pfeifer__cppnow_05-19-2017.pdf
+
+cmake_minimum_required(VERSION 2.8)
 
-set(SimGrid_PATH ${SimGrid_PATH} CACHE PATH "Path to SimGrid library and include")
 find_path(SimGrid_INCLUDE_DIR
   NAMES simgrid_config.h
-  HINTS ${SimGrid_PATH}/include /usr/include /usr/local/include /opt/simgrid/include
+  PATHS /opt/simgrid/include
```
mquinson: Why are you removing the hint on ${SimGrid_PATH}/include here? When SimGrid is installed in non-standard location (what some people want or must do), this environment variable allows to let cmake find it anyway.

FindSimGrid.cmake
```diff
@@ -1,60 +1,92 @@
...
+)
 
+if (SimGrid_FOUND AND NOT CMAKE_VERSION VERSION_LESS 2.8.12)
+  add_library(SimGrid::SimGrid SHARED IMPORTED)
+  set_target_properties(SimGrid::SimGrid PROPERTIES
+    INTERFACE_SYSTEM_INCLUDE_DIRECTORIES ${SimGrid_INCLUDE_DIR}
+    INTERFACE_COMPILE_FEATURES cxx_alias_templates
+    IMPORTED_LOCATION ${SimGrid_LIBRARY}
+  )
+  # We need C++11, so check for it
```
mquinson: This is kinda useless, as SimGrid will simply not compile without C++11. No need to test for it when SimGrid is found.

(unless I'm missing something, maybe)

FindSimGrid.cmake
```diff
@@ -1,60 +1,92 @@
...
 
-# Extract the actual version number
-file(READ "${SimGrid_INCLUDE_DIR}/simgrid_config.h" _SimGrid_CONFIG_H_CONTENTS)
-set(_SimGrid_VERSION_REGEX ".*#define SIMGRID_VERSION_STRING \"SimGrid version ([^\"]*)\".*")
-if ("${_SimGrid_CONFIG_H_CONTENTS}" MATCHES "${_SimGrid_VERSION_REGEX}")
-      string(REGEX REPLACE "${_SimGrid_VERSION_REGEX}" "\\1" SimGrid_VERSION "${_SimGrid_CONFIG_H_CONTENTS}")
-endif ("${_SimGrid_CONFIG_H_CONTENTS}" MATCHES "${_SimGrid_VERSION_REGEX}")
-  
+  # Concat the matches to MAJOR.MINOR.PATCH assuming they appear in this order
+  foreach(match ${SimGrid_VERSION_STRING})
```
mquinson: I would have prefered to use SIMGRID_VERSION in the same file, which is defined to something like:
#define SIMGRID_VERSION (100UL * (100UL * (3) + (18)) + (90))

That's generated automatically, so maybe we can come up with the right regexp. But that's just a picky comment, feel free to not address this one.

FindSimGrid.cmake
Flamefire: The question is: How is this variable set? By the user?

The preferred (aka CMake-)way is to use the appropriate CMake variables: https://cmake.org/Wiki/CMake_Useful_Variables
So that would be `CMAKE_PREFIX_PATH` or `CMAKE_INCLUDE_PATH`/`CMAKE_LIBRARY_PATH`

We have been doing the same with a `Foo_PATH` before till someone noted, that it is non-standard. Although "standard" is pretty undefined for CMake. So I can add it again if you want. Just wanted to point out that there is a "CMake-way" of pointing to externals.

Things are different, if you would set SimGrid_PATH from something like the install script or so.

FindSimGrid.cmake
Flamefire: True. But this is called **after**  Simgrid was compiled and installed. So not in the same context as the SimGrid compilation.

Just a quick outline of the use-case: Compile SimGrid and install it to e.g. /usr/local. Place FindSimGrid.cmake somewhere where user projects can find it (`install(EXPORT` and similar do that automatically if one was to use that). End of SimGrid context! Now a user comes with a fresh CMakeLists.txt for his projects and calls `find_package(SimGrid)`. Nothing is guaranteed about what the user is using. Could be C++17, C++98, ...

So yes, we need to put C++11 as a (minimum) requirement for the installed SimGrid just the same as you would put library dependencies there.

FindSimGrid.cmake
Flamefire: I would suggest to do it the other way round: Calculate SIMGRID_VERSION from those defines above to remove the redundancy. Or calculate the other values from this version.

I chose to use the MAJOR/MINOR/PATCH as they are available in all SimGrid versions. And you might change SIMGRID_VERSION to the value directly (e.g. 31890) which would not break (C++)Code but the CMake RegExp.

Well in the end: Choose your poison and tell me which one you'd like ;)

FindSimGrid.cmake
mquinson: The whole thing is that in my mind, FindSimGrid could/should be embeeded in applications relying on SimGrid. If not, there is no way to find simgrid when it's not in the regular path. And if it is (if simgrid is installed in /usr), there is no need for a FindSimGrid since we don't do anything complicated.

So, I need the hint to allow users compiling applications that rely on SimGrid on machines where SimGrid is not in the default path (eg in ~ instead of /usr). Most people wont use the HINT, but removing it will make the life of BatSim's users more difficult for very little gain on our side.

Likewise, I don't see the point of testing C++11 from FindSimGrid because you will principally need this macro file on systems which non-standard installation of SimGrid. IE probably on systems where you compiled simgrid yourself (and if so, then C++11 is there for sure). 
If you installed simgrid from a package (like the debian one), you will most probably not use FindSimGrid but use the package manager for dependency handling.

That being said, I am not opposed to the testing of C++11 if it's important to you. As long as my envisionned use-case (where you copy FindSimGrid in the depending projects) is possible, I'm ok with including the other use-case (where cmake is used as a sort of pkg-config). Simply, you'll probably also have to patch the installation targets because I'm not sure of how we should install that file for cmake to easily find it.

The only point that is very important to me is the hint thing. That's mandatory for my use-case. Would you agree on this?

FindSimGrid.cmake
mquinson: Of course, SIMGRID_VERSION is automatically generated.
https://github.com/simgrid/simgrid/blob/master/include/simgrid_config.h.in#L30

FindSimGrid.cmake
mquinson: Just to complete, I propose to make it possible to use FindSimGrid in the same way than we are using other modules: https://github.com/simgrid/simgrid/tree/master/tools/cmake/Modules

FindSimGrid.cmake
Flamefire: Yes, from a CMake point of view. However what opposes replacing  
`#define SIMGRID_VERSION (100UL * (100UL * (@SIMGRID_VERSION_MAJOR@) + (@SIMGRID_VERSION_MINOR@)) + (@SIMGRID_VERSION_PATCH@))` with
`#define SIMGRID_VERSION (100UL * (100UL * (SIMGRID_VERSION_MAJOR) + (SIMGRID_VERSION_MINOR)) + (SIMGRID_VERSION_PATCH))` ?

This makes the origin of the numbers clear to the user (who only has the `.h` file)

mquinson:
> Yes, from a CMake point of view. However what opposes replacing  
> `#define SIMGRID_VERSION (100UL * (100UL * (@SIMGRID_VERSION_MAJOR@) + (@SIMGRID_VERSION_MINOR@)) + (@SIMGRID_VERSION_PATCH@))` with
> `#define SIMGRID_VERSION (100UL * (100UL * (SIMGRID_VERSION_MAJOR) + (SIMGRID_VERSION_MINOR)) + (SIMGRID_VERSION_PATCH))` ?

Done locally, but not pushed to save one recompilation cycle to the
build daemons.

FindSimGrid.cmake
Flamefire: I think you might be missing a point of the FindModules: They are to be used like `pkg-config`! Because what else is it doing? It gives you the compile and link flags for an external dependency. The difference is where those are set. You can either have it in CMake (`FindBoost`), in the installation (e.g. in `<prefix>/lib/cmake/<name>`, config-only) or somewhere where `CMAKE_MODULE_PATH` points to.

So directly on your comments: `FindSimGrid` is supposed to be embedded in applications using it. Or at least be somewhere, where `CMAKE_MODULE_PATH` points to.   
It then searches for Simgrid in the default locations, that is `/usr`, `/usr/local`, `/usr/local/lib/<arch>`. So here you already need the find module as you don't know, **what** the *default path*  is. (yes, you could simply add `-lSimGrid` to the linker flags and hope, that the library and the includes are in the compiler default paths, but you usually want a failure during configure time not compile time and you would also not have version checking)

So about the hint: Your use case is, that the user has installed Simgrid into a location not searched by CMake by default. He already has access to `FindSimGrid` (e.g. he copied it from the repo to his app). Now you want this user to be able to do: `export SimGrid_PATH="..."; cmake .` Correct?   
Well they still can do this: `export CMAKE_PREFIX_PATH="...:$CMAKE_PREFIX_PATH"; cmake .` The cool thing about this is, that it works for every library. There is no need to look at the source of any find-module to search if it has `Foo_PATH`, or how it is spelled (I know others using e.g. `Foo_DIR`, `FOO_DIR`, `FOO_ROOT`, etc.) So this approach is more flexible and scalable. Some of the projects I was working on moved to that approach with this reasoning and I find it very convincing.

C++11: Again: You also need it for "standard installations" (see above), especially when you decide to add further requirements (maybe a new external dependency?)  And yes "C++11 is there", but it might not have been enabled. Remember that you need to pass `-std=c++11` explicitly to the compiler. So without this C++11 dependency a user won't be able to compile his application **using** SimGrid unless he sets that flag in **his app**.   
Comparing example: Imagine a static library `libFoo` that requires another one `libBar`. The Find-Module is expected to include this dependency. So (in old CMake) you would get `FOO_LIBRARIES="foo bar"`.   
The C++ standard is the same: It is just another dependency. Newer CMakes handle that very well.

2 Examples: A user has SimGrid built and installed (e.g. by the admin or package manager) and wants to use that by simply doing a `find_package` with `target_link_libraries(... SimGrid::SimGrid)`.

1. He uses only C++98 in his code, so no C++11 mentioning in his CMakeLists.txt. CMake detects that and automatically adds the C++11 flag to the compiler flags.
2. He uses C++17 by setting that in his CMakeLists.txt. CMake detects that and does not add anything since it knows that C++11<=C++17

Package Manager: Yes the package manager takes care about runtime dependencies. But it doesn't take care about compile time dependencies. The include path is a compiler flag: `-I<SIMGRID_INCLUDE>`, the library a linker flag `-L<SIMGRID_LIB_DIR> -lsimgrid` and the C++11 likewise a compiler flag `-std=c++11`. All of them are clearly out of the scope of any package manager (unless they stuff the includes and libraries right into standard accessible paths, but this will also fail in some cases)

Summary: Your use case is fully supported. In fact your use case is the only one actually supported. It properly records all required dependencies. Example `CMakeLists.txt`: https://gist.github.com/Flamefire/2cce7a3df13ee2602c4044ac00600bf6 I think this example makes it clear, how the new Find-Module can be used and how this makes the library-uses easier.

Final note: Nothing of this is important to me personally. If you don't want that C++11 dependency adding, I'll remove it. If you want the `SimGrid_PATH` variable (especially if this is set by some tool or so which I don't know of), I'll re-add it. If any of my reasoning is wrong, please correct me. My motivation for this PR is to improve my CMake-Fu and practice using the "more modern" features to simplify usages. So I hope you don't get this as an argument but a discussion which should be advantageous for everyone.

Flamefire:
I found an interesting detail in https://cmake.org/cmake/help/v2.8.12/cmake.html#command:find_package : When switching to the `CONFIG` mode for packages (basically install a `SimGridConfig.cmake` and `SimGridVersion.cmake` with the library instead of providing a Find-Module) then CMake will search for such a `*Config.cmake` in some possible prefix dirs. Those include `CMAKE_PREFIX_PATH` (either as a CMake cache variable or environment variable) and `<package>_DIR` (in this case `SimGrid_DIR`) environment variable. So using the `CONFIG` mode would allow:

- Use a custom FindSimGrid.cmake if required (e.g. for old versions) which takes precedence over CONFIG mode
- Use the scalable `CMAKE_PREFIX_PATH`
- Use the (older) `SimGrid_DIR` (note the `DIR` not `PATH`)

Setting this up seems to be pretty easy (about a dozen lines added to current CMakeLists.txt to create and install these files) but requires CMake >=2.8.11 to build SimGrid.

Downside: There is no new Find-Module hence users of older SimGrid versions must default to the old Find-Module or you distribute this PR as a replacement (as it is more powerful and would even be compatible to the CONFIG mode, so great for transitioning)

PR #272, opened 2018-06-21
Add TIT trace conversion script

mickours:


agiersch:
Hi,

Make distcheck needs to be fixed.
You should also add a rule to install this script.

I think that it should be done in tools/CMakeLists.txt and tools/cmake/Distrib.cmake

Arnaud

mickours:
Hi Arnaud,

I've added the script in the `bin_files` list inside `tools/CMakeLists.txt`, but I'm not sure of what you want me to do with distcheck, can you give me more details?

agiersch:
Hi,
Just run "make distcheck", and see if it completes without error.
This target checks that the archive generated by "make dist" is complete and does not embed unwanted files.

PR #279, opened 2018-06-29
Fix breakpoint option

sthibaul:
It was using relative time deltas instead of absolute time.


mquinson:
whoops :) 
Thanks, Sam.

PR #283, opened 2018-07-06
Smpi convert script fixes

mickours:
- Add an assert to avoid rewriting traces files if absolute path are used in the trace list file.
- Make everything lowercase in TI traces


mquinson:
Some documentation would be welcome in that script ;)

PR #292, opened 2018-08-12
Cover with a test Mailbox::ready() method introduced in commit 1ed0e64dc40

kovin:


mquinson:
Please allow some time for us to integrate it. I merged a Merge Request on framagit.org/simgrid that was not ready yet, and we try to fix it before we integrate other contributions. 

It would have been easier if I did not integrate it too early in the first place, sorry.

PR #302, opened 2018-09-25
Rename s4u synchro examples

mpoquet:
Rename:
- example `s4u-barrier` into `s4u-synchro-barrier`
- example `s4u-mutex` into `s4u-synchro-mutex`

cmake/ctest agree with the modification on my machine and sphinx's output seems fine.

NB: this PR is only the last commit. Github's master is a few commits behind gforge's master and I cannot push on Github.


mquinson:
thanks a lot!

PR #303, opened 2018-10-01
S4u semaphore

mpoquet:
Implements `s4u::Semaphore` as a direct wrapper around SIMIX and tests it.


mquinson:
Hello, thanks for this change, I merged it. 

I'm not sure why you're using a mutex in your example. Semaphore should work on their own, not like condition variable... But that's not a big deal, thanks for your work.

PR #311, opened 2018-11-12
argc, argv -> explicit parameters

mpoquet:
### Motivation
The ``smpi-replay-multiple-manual`` examples leaked memory which could not be avoided without a dirty hack in the example, requiring the example to know deep implementation details of the library.

After some investigation the argc/argv hack was deep within SMPI and required (small) modification in several parts of the code base.

### Impact overview
Pros
- Way clearer smpi-replay API
- SMPI made a bit more modular, which may be required soon to reuse its code from OpenMPI

Cons
- Reading of argc/argv is not centralized deeply anymore, the external interface**s** (online/offline) are in charge of handling them.

### Content
This PR decreases the usage of argc/argv to use explicit parameters instead — removing the argc/argv hacks in the replay part of the library.

As the argc/argv hack was in the common part of SMPI (smpi actors), this update also required modification to the code in charge of online SMPI simulation.

Most smpi-replay examples have been slightly modified because of the API change of this part.

Online SMPI examples have not been modified but ``tesh-smpi-coll-allreduce`` — just to adapt argc/argv as the example uses them.

### DIFF
--- a/examples/s4u/replay-comm/s4u-replay-comm.cpp
+++ b/examples/s4u/replay-comm/s4u-replay-comm.cpp
@@ -30,16 +30,9 @@ class Replayer {
 public:
   explicit Replayer(std::vector<std::string> args)
   {
-    int argc;
-    char* argv[2];
-    argv[0] = &args.at(0)[0];
-    if (args.size() == 1) {
-      argc = 1;
-    } else {
-      argc    = 2;
-      argv[1] = &args.at(1)[0];
-    }
-    simgrid::xbt::replay_runner(argc, argv);
+    const char* actor_name     = args[0].c_str();
+    const char* trace_filename = args[1].c_str();
+    simgrid::xbt::replay_runner(actor_name, trace_filename);
   }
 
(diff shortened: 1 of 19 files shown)


mpoquet:
Mmh they will be affected with the current state of the PR.
I'll redo the [previous trick](https://github.com/simgrid/simgrid/pull/311/commits/688a1ca4701a91d083af47d96a146d5c223ee165#diff-7f8b70f074dda42c242618f30ae565d1L72) to remove first positional parameters and update the PR.

mpoquet:
I do not know if [making mutable argc/argv for FORTRAN](https://github.com/simgrid/simgrid/blob/d7809554918bc1cf4825e9718fc61649ae6f0ed5/src/smpi/internals/smpi_global.cpp#L437) is still needed now that ``simgrid::smpi::ActorExt::init`` does not change its argument, what do you think @adegomme?

mpoquet:
Opening another PR for the clean-up.

PR #315, opened 2018-11-20
replay from smpirun (without a dedicated replayer program)

mpoquet:
# **This PR is based on #314, please merge it first.**

### What?
This PR adds the `smpirun [...] -replay <tracefile> [program] [program-options]` syntax to smpirun.
- Allows to replay a tracefile without specifying a replay program (a default one is used)
- Allows to specify a replay program if desired (used by the `smpi-ampi` example for example)

This has led to the following.
- Changed the `smpirun` usage (removal of `-ext`, introduction of `-replay <tracefile>`, `program` is optional if `-replay` is used).
- Update of the replay tests that were using `smpirun`.
- Introduction of a new `smpi-replay` test for the default replayer (`smpireplaymain`).
- Renaming of the previous `smpi-replay` test to `smpi-replay-override-replayer`.

### How?
This is done by creating a new `smpireplaymain` executable, which is used by default by `smpirun` if no program is given. This program is installed next to `smpimain` (the `${PREFIX}/lib` location is strange but that's for another issue).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -426,7 +426,7 @@ if(CMAKE_USE_PTHREADS_INIT)
             RUN_OUTPUT_VARIABLE semopen_runmsg)
     
     if(semopen_compilable)
-      if(NOT semopen_retval) # error if not 0
+      if(NOT semopen_retval) # error if not 0
         message(STATUS "sem_open is compilable and executable")
 	set(HAVE_SEM_OPEN 1)
       else()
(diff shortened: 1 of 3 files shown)


mquinson:
I was kinda hopping that this would be possible without adding a new binary. But maybe it would make the smpi_main binary hairly and your approach is preferable. I'm not completely sure yet. Let's merge #314 first.

mquinson:
the smpi-replay test is failing on travis with the following error message:
```
+[0.000000] [xbt/CRITICAL] dlopen failed for /home/travis/build/simgrid/simgrid/lib/simgrid/smpireplaymain: /home/travis/build/simgrid/simgrid/lib/simgrid/smpireplaymain: cannot dynamically load executable (errno: 0 -- Success)
```
I'd like to understand why before merging. I'm puzzled by the fact that the tests pass on circleci.

agiersch:
Tests smpi-replay and smpi-replay-override-replayer may fail when run concurrently
`ctest -R 'smpi-replay$|smpi-replay-override' -j2`

PR #330, opened 2019-10-02
Minor polishing of English: Configuring_Simgrid.rst

gc00:
The PR has some minor changes to the English.

Separately, I have some recommended formatting changes that I wouldn't know how to submit:   All of this refers still to the file `Configuring_Simgrid.rst`.   Perhaps someone else can consider those changes:

```
The rst file says:  "as formatted by the ltl2ba program"
[ If this is destined for the web, can we include a link from ltl2ba program
  to:  https://github.com/utwente-fmt/ltl2ba
  or else include the URL in parentheses?
  As a new user, i had had trouble tracking down this pointer. ]

consumption of the snapshots to be @f$ @mbox{number of processes}
@times @mbox{stack size} @times @mbox{number of states} @f$.
[ Fix the formatting?  It didn't display properly on the web. ]

I don't understand this sentence:
"The other options should be the same (but the model-checker should be disabled)."
from the section: "Replaying buggy execution paths out of the model-checker"
    This seems to mean that you should replay using simgrid-mc (the model checker),
but you should also disable the model checker.  Could someone add a bit mroe explanation here?

using @c “no” as a value.
[ Fix the formatting?  It didn't display properly on the web. ]   
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,6 +4,13 @@ set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} ${CMAKE_HOME_DIRECTORY}/tools/cmake/M
 
 project(SimGrid C CXX)
 
+## This is a better way to set the flags:
+if(NOT CMAKE_BUILD_TYPE)
+  set(CMAKE_BUILD_TYPE Release)
+endif()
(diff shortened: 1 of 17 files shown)


mquinson:
Since this commit was also integrated in #332 , this is already merged. I'm leaving this PR open because the other items, given in the text here, are not fixed yet.

PR #331, opened 2019-10-04
Fix a doc error about actors (Tutorial_algorithms)

Webcretaire:
In the "[Simulating Algorithms](https://simgrid.org/doc/latest/Tutorial_Algorithms.html)" section of the docs, a master / worker application is used as an example. I think that when presenting actors a "worker" has been written in place of "master"


mquinson:
Obviously. Thanks for taking the time to report it, and for your interest in SimGrid.

PR #332, opened 2019-10-04
Add '-O0' flag for: cmake -DCMAKE_BUILD_TYPE=Debug

gc00:
Currently, when using cmake as:
```
cmake -DCMAKE_BUILD_TYPE=Debug
```
the g++ flags get set as something like "`-g -O2`".  This makes it difficult to use GDB.  This commit will add the `-O0` flag after the `-O2` in the `Debug` case, to ensure easier debugging.

Feel free to modify my comment in the commit, if you like.


mquinson:
Thanks for this.

Issue #189, opened 2017-06-15
Energy consumption of parallel tasks


mpoquet:
Hello,

I just found out that that hosts computing parallel tasks do not consume more energy than when they are idle. Is this behaviour desired? Are my platform and my code alright?

I used computation-only parallel tasks, which I thought would completely use the host computational resource.

MWE gist [there](https://gist.github.com/mpoquet/8dfcb26671415a5c4adec325381a3f8c) (short C++ file, platform XML file and a Makefile).

mpoquet:
Patch with @shurakai:
``` patch
--- a/src/surf/plugins/host_energy.cpp
+++ b/src/surf/plugins/host_energy.cpp
@@ -207,8 +207,8 @@ double HostEnergy::getCurrentWattsValue(double cpu_load)
   double power_slope   = 0;
 
   if (cpu_load > 0) { /* Something is going on, the machine is not idle */
-    double min_power = range.min;
-    double max_power = range.max;
+    min_power = range.min;
+    max_power = range.max;
 
     /**
      * The min_power states how much we consume when only one single
@@ -220,15 +220,18 @@ double HostEnergy::getCurrentWattsValue(double cpu_load)
      * i.e., we need min_power + (maxCpuLoad-1/coreCount)*power_slope == max_power
      * (maxCpuLoad is by definition 1)
      */
-    double power_slope;
     int coreCount         = host->coreCount();
     double coreReciprocal = static_cast<double>(1) / static_cast<double>(coreCount);
     if (coreCount > 1)
+    {
       power_slope = (max_power - min_power) / (1 - coreReciprocal);
+      current_power = min_power + (cpu_load - coreReciprocal) * power_slope;
+    }
     else
-      power_slope = 0; // Should be 0, since max_power == min_power (in this case)
+    {
+      current_power = max_power;
+    }
 
-    current_power = min_power + (cpu_load - coreReciprocal) * power_slope;
   } else { /* Our machine is idle, take the dedicated value! */
     current_power = range.idle;
   }
```

mpoquet:
Looks like the patch (commit e96681fb8) fixed the problem.

![energy_plot](https://user-images.githubusercontent.com/3502831/27192715-9f7191f8-51fc-11e7-8699-c828e76cef70.png)

Shurakai:
[...]

PR #334, opened 2019-10-08
Rework energy plugin, again

Mommessc:
PR related to last discussion on #189.

### DIFF
--- a/include/simgrid/plugins/energy.h
+++ b/include/simgrid/plugins/energy.h
@@ -15,6 +15,7 @@ XBT_PUBLIC void sg_host_energy_plugin_init();
 XBT_PUBLIC void sg_host_energy_update_all();
 XBT_PUBLIC double sg_host_get_consumed_energy(sg_host_t host);
 XBT_PUBLIC double sg_host_get_idle_consumption(sg_host_t host);
+XBT_PUBLIC double sg_host_get_idle_consumption_at(sg_host_t host, int pstate);
 XBT_PUBLIC double sg_host_get_wattmin_at(sg_host_t host, int pstate);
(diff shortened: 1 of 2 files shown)


include/simgrid/plugins/energy.h
```diff
@@ -14,7 +14,7 @@ SG_BEGIN_DECL()
 XBT_PUBLIC void sg_host_energy_plugin_init();
 XBT_PUBLIC void sg_host_energy_update_all();
 XBT_PUBLIC double sg_host_get_consumed_energy(sg_host_t host);
-XBT_PUBLIC double sg_host_get_idle_consumption(sg_host_t host);
+XBT_PUBLIC double sg_host_get_wattidle_at(sg_host_t host, int pstate);
```
mquinson: I think that `sg_host_get_wattidle_at()` should exist in addition to `sg_host_get_idle_consumption()` since they don't really cover the same goal. `sg_host_get_idle_consumption()` is evaluating it at the current pstate.

In addition, if you want to remove a function from the public interface, you should deprecate the old one without removing it for one full year. 

Finally, `sg_host_get_wattidle_at()` should probably be named `sg_host_get_idle_consumption_at()` for sake of consistency.

Mommessc: Actually `sg_host_get_idle_consumption()` calls `get_idle_consumption()` that simply retrieve the idle consumption of the first pstate [see here](https://github.com/simgrid/simgrid/blob/master/src/plugins/host_energy.cpp#L230),
that's why I wanted to replace it.

For the second point, I think it's better to keep consistency with the other functions `sg_host_get_wattmin_at()` and `sg_host_get_wattmax_at()` since they are getters of similar values.

PR #344, opened 2020-07-08
cast hugepages macros parameters to int64

llnns:
The two macros `ALIGN_UP` and `ALIGN_DOWN` used by the shared malloc hugepages gave the wrong result in a logical operation because of a different precision type.

The following minimal example ilustrates this:

```C++
#include <stdio.h>
#include <cstdint>

#define ALIGN_UP(n, align) (((n) + (align)-1) & -(align))
constexpr unsigned HUGE_PAGE_SIZE = 1U << 21;

int main(){

  void* mem;
  void* allocated_ptr = (void*)0x7ffdffef1000;

  mem = (void*)ALIGN_UP((int64_t)allocated_ptr, HUGE_PAGE_SIZE);
  printf("ap: %p\n", allocated_ptr);
  printf("mem: %p\n", mem);

}
```
The value `0x7ffdffef1000` is returned from `mmap` in simgrid code, but when `ALIGH_UP` is applied, `mem` is `NULL`.
The `align` parameter should be of the same type of `n` before the `-(align)`.

Some of the calls were already casting the second parameter.

### DIFF
--- a/src/smpi/internals/smpi_shared.cpp
+++ b/src/smpi/internals/smpi_shared.cpp
@@ -184,7 +184,7 @@ void* smpi_shared_malloc_partial(size_t size, size_t* shared_block_offsets, int 
                                 "to allow big allocations.\n",
              size >> 20);
   if(use_huge_page)
-    mem = (void*)ALIGN_UP((int64_t)allocated_ptr, (int64_t)HUGE_PAGE_SIZE);
+    mem = (void*)ALIGN_UP((int64_t)allocated_ptr, HUGE_PAGE_SIZE);
   else
     mem = allocated_ptr;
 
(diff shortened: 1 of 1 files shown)


mquinson:
Thanks for this very interesting patch. It must have been a nightmare to debug :)

But wouldn't it be better to fix the definition of the `ALIGN_UP` and `ALIGN_DOWN` macros instead of fixing all calling location? Maybe something as follows, even if it's not really readable :

```
#define ALIGN_UP(n, align) (((((int64_t)n) + ((int64_t)align)-1) & -((int64_t)align))
#define ALIGN_DOWN(n, align) (((int64_t)n) & -((int64_t)align))
```

llnns:
> But wouldn't it be better to fix the definition of the `ALIGN_UP` and `ALIGN_DOWN` macros instead of fixing all calling location? Maybe something as follows, even if it's not really readable :
> 
> ```
> #define ALIGN_UP(n, align) (((((int64_t)n) + ((int64_t)align)-1) & -((int64_t)align))
> #define ALIGN_DOWN(n, align) (((int64_t)n) & -((int64_t)align))
> ```

Yeah, that really looks better, the new commit does it.

Thanks

Issue #346, opened 2020-09-01
[SMPI] error while loading shared libraries: libsimgrid.so


Ezibenroc:
With the version from `master` branch, when calling `smpirun`, I get the following error:
```
/usr/local/lib/simgrid/smpimain: error while loading shared libraries: libsimgrid.so.3.25.1: cannot open shared object file: No such file or directory
```
It works fine with the last release. I tried some git bisect to find the culprit, but some commits introduced other bugs, so this is not straightforward.
What I can tell:
- this issue exists since at least 094f8c02076206ad96ec3bbad1af2d387534a002
- with the older commit 10759b01e9fac33c832b716a0fbe8c2c4e7e8859 I could not even call `smpicc`
- with the even older commit 50c10d53194359944630b389b70f7530aa63d277, this worked as expected

------

Here is a minimal example to reproduce the issue.

Take the following file `test.c`:
```c
#include <mpi.h>
#include <stdio.h>

int main(int argc, char** argv) {
    MPI_Init(&argc, &argv);
    int world_size;
    MPI_Comm_size(MPI_COMM_WORLD, &world_size);
    int world_rank;
    MPI_Comm_rank(MPI_COMM_WORLD, &world_rank);

    int buff = world_rank*27+12;
    if(world_rank % 2 == 0) {
        printf("rank %d sends    %d to rank %d\n", world_rank, buff, world_rank+1);
        MPI_Send(&buff, 1, MPI_INT, world_rank+1, 0, MPI_COMM_WORLD);
    }
    else {
        MPI_Recv(&buff, 1, MPI_INT, world_rank-1, 0, MPI_COMM_WORLD, NULL);
        printf("rank %d receives %d from rank %d\n", world_rank, buff, world_rank-1);
    }
    MPI_Finalize();
}
```

This file `platform.xml`:
```xml
<?xml version="1.0"?>
<!DOCTYPE platform SYSTEM "http://simgrid.gforge.inria.fr/simgrid/simgrid.dtd">

<platform version="4.1">
    <AS id="AS0" routing="Full">
        <cluster id="cluster" prefix="pc-" suffix="" radical="0-1" speed="1" bw="12.5GBps" lat="0.10us"/>
    </AS>
</platform>
```

And this file `hosts.txt`:
```
pc-0
pc-1
```

Compile and run:
```bash
smpicc test.c -o test
smpirun --cfg=smpi/privatize-global-variables:dlopen --cfg=smpi/display-timing:yes  \
[...]

PR #347, opened 2020-09-01
[SMPI] Fix shared-malloc with huge pages

Ezibenroc:
After updating my Simgrid installation to the latest stable release, I found out that I could not perform a shared-malloc with huge pages, I get the following error:
```
[dahu-13.grid5000.fr:401:(402) 0.007487] /tmp/simgrid/src/smpi/internals/smpi_shared.cpp:275: [root/CRITICAL] Could not map folded virtual memory (Cannot allocate memory). Do you perhaps need to increase the size of the mapped file using --cfg=smpi/shared-malloc-blocksize:newvalue (default 1048576) ?You can also try using  the sysctl vm.max_map_count
```

The problem has been introduced by b024b9e443552c94609fc56a240cdb199a329853

The constant `HUGE_PAGE_SIZE` needs to be a 64 bit integer, otherwise we get an integer overflow in [smpi_shared.cpp](https://github.com/simgrid/simgrid/blob/master/src/smpi/internals/smpi_shared.cpp#L187):
```c++
    mem = (void*)ALIGN_UP(allocated_ptr, HUGE_PAGE_SIZE);
```

**Note** that I could not test this pull request with the `master` branch due to the bug reported in #346. I tested it with v3.25 and it worked as expected.

### DIFF
--- a/src/smpi/internals/smpi_shared.cpp
+++ b/src/smpi/internals/smpi_shared.cpp
@@ -145,8 +145,8 @@ static void *smpi_shared_malloc_local(size_t size, const char *file, int line)
 #define ALIGN_UP(n, align) (((int64_t)(n) + (int64_t)(align) - 1) & -(int64_t)(align))
 #define ALIGN_DOWN(n, align) ((int64_t)(n) & -(int64_t)(align))
 
-constexpr unsigned PAGE_SIZE      = 0x1000;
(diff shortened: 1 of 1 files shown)


agiersch:
Isn't it already fixed by commit f46e2b4682b6bd82813d92604d698781f9bb34df?

Ezibenroc:
Indeed, I did not spot this commit, and as I said I could not test the `master` branch.  Thanks!

PR #350, opened 2020-11-02
pull request 2

eazimi:
At first, I decided to put all parts of state class in the façade layer. I changed my mind because I thought a state is a proprietary characteristic for a model checker algorithm. I decided to follow the following approach:
1. The State can contain some information that comes from kernel part, but it is banned from getting this information directly from the kernel segment. The State can retrieve information by calling specific APIs of façade layer.
2. Some fields the represent some information originated from kernel segment are moved to façade layer. The corresponding APIs are provided by façade layer to give the capability to the State instance to get required information from façade layer.
The mc_state_dev and state_detail classes aim to implement this approach.


agiersch:
Most commits of this PR are already merged through commit 6b8d58b226c5c14d9b27d7ba55230d7ba7cfbd16.
Moreover it seems superseded by PR #351.

PR #366, opened 2020-11-27
More updates in comm. deter. checker to let it to use APIs of the facade layer

eazimi:
Developments in comm. deter. checker


agiersch:
This PR can be closed. Its content is already merged through commit 6b8d58b226c5c14d9b27d7ba55230d7ba7cfbd16.

PR #382, opened 2022-09-28
Ensure a final fPIC in case CMDARGS has a fPIE

llnns:
This changes the order of the CMDLINE to place the CFLAGS after CMDARGS.

Useful when CMDARGS has an fPIE (coming from a CMake POSITION_INDEPENDENT_CODE, for example), this will ensure a fPIC that may be present in CFLAGS to be after the fPIE in CMDARGS.

### DIFF
--- a/src/smpi/smpicc.in
+++ b/src/smpi/smpicc.in
@@ -93,6 +93,10 @@ while [ $# -gt 0 ]; do
     esac
 done
 
+if [ "x@WIN32@" != "x1" ]; then
+    list_add CMDARGS "-fPIC"
+fi
+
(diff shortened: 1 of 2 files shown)


agiersch:
I guess that it will break if the user needs to override other flags. For example "-std=..." with smpicxx

llnns:
Ok, maybe now adding a fPIC at the end of the CMDARGS?

agiersch:
It looks better like that, thanks.
Is there any reason not to remove the lines appending -fPIC to CFLAGS/CXXFLAGS?

llnns:
I don't think so, removed as well

Issue #389, opened 2023-12-19
Routing algorithm is broken when using 3+ level nested zones


JRodez:
Hi there, 

# Context
> # tl; dr : clarifying my use case
> I need to represent a complex hyerarchical network with far more than 2 nested zones via the C API.

For my PhD thesis I'm working in a Fog architecture simulation using Batsim.
My platform contains machines composed of multiple GPU/CPU, which means each machine is a StarZone with hosts. 
## Overview of my platform :
![sgp_het_fog](https://github.com/simgrid/simgrid/assets/80471345/bb6f48e1-a6bc-496d-8b16-44f9da179a14)

As my platform is heavy heterogeneous and has many levels of Zones, I think my work can contribute, some ways, to this fantastic library.

# Routing issue
It seems that the routing algorithm does not work for many recursive subzones ( more than 3 included ).
I made a simple scenario example where this problem occurs, using [this flatify program](https://gist.github.com/JRodez/e5f3b22fdecaa9285c1fb905e479b330) over 
[a 3 level platform](https://gist.github.com/JRodez/2a4645edb045481557af0facb4e38ecb#file-routing_cluster_bug-cpp):

Overview of the log [clusterbugged.log](https://github.com/simgrid/simgrid/files/13719487/clusterbugged.log) :

> ```
> [0.000000] ./src/xbt/exception.cpp:48: [xbt_exception/CRITICAL] Uncaught exception std::out_of_range: unordered_map::at
> [0.000000] ./src/xbt/exception.cpp:75: [xbt_exception/CRITICAL] Current backtrace:
>   ->  #0 0x00007F3558AEFBC7 in /lib/simgrid_lib/lib64/libsimgrid.so.3.35.1
>   ->  #1 0x00007F35586B4F8A in /nix/store/myw67gkgayf3s2mniij7zwd79lxy8v0k-gcc-12.3.0-lib/lib/libstdc++.so.6
>   ->  #2 0x00007F35586B4FF5 in /nix/store/myw67gkgayf3s2mniij7zwd79lxy8v0k-gcc-12.3.0-lib/lib/libstdc++.so.6
>   ->  #3 0x00007F35586B5247 in /nix/store/myw67gkgayf3s2mniij7zwd79lxy8v0k-gcc-12.3.0-lib/lib/libstdc++.so.6
[...]

PR #392, opened 2024-01-23
New route resolving algorithm for issue #389

JRodez:
Hi there,

According to #389  I propose to integrate this algorithm as an alternative, currently not by replacing the existing algorithm, but by enabling the macro `HIGH_DEPTH_ROUTING_ALGORITHM`. 

This new algorithm is purely sequential, since the route can be resumed to the concatenation of the route from the source to the common ancestor and the route from the common ancestor to the destination.

## The Algorithm 
So the algorithm is vulgarly the following  :
1) Fill a vector (i call it fullvector for explanation)  with all zones between the rootzone and the netpoint for the two subroutes.
2) Find the common ancestor of the two netpoints and the corresponding first zones of each netpoints under the common_ancestor. We remove the part of the fullvector from the rootzone to the common ancestor included.
3) Calculate the route between the source ancestor and the destination ancestor, which gives us the gateways for computing the source to source ancestor route and the destination ancestor to destination route
4) for the two subroutes and according to their direction : for each zone of fullvector, compute the route between the gateway and the zone netpoint, which will give us the gateway for this zone for the next iteration. Insert the links in a vector according to the direction.
5) Calculate the route between the gateway and the netpoint.
6) Add the two subroutes to the route.

## Performance 
There are some problems with the current recursive algorithm, which I explained in #389. 
For me, another problem is that find_common_ancestor is called for each pair of zones, but it [computes the path between them to rootzone](https://github.com/JRodez/simgrid/blob/ddafba460da954dcb90898b4e91bd90ac177bca1/src/kernel/routing/NetZoneImpl.cpp#L723-L735) for each subroute in the final route.

This means that for a platform with a level of zones from root to hosts of 4, the current algorithm computes 4 times the subparts of the same vector 2 times (src and dst ) when rootzone is the first common ancestor of src and dst.

### Comparing complexity

So to compute routes between 2 points (with root zone as first common ancestor) for a $d$ platform depth (in terms of zones), we regenerate parts of the same 2 "host to root zone zones" vectors $`O(2*d)`$ times.
The new algo just generates the 2 vectors 1 time, removing the common parts of them, and iterates on them 1 time for good and all. In the worst case it costs $`O(2)`$.

### Measurements
I measure a performance gain of up to 10% to 30% using perf on my highly nested zones with a [flatify program](https://gist.github.com/JRodez/e5f3b22fdecaa9285c1fb905e479b330).
I don't have much time to investigate, but I'm pretty sure that this algorithm has a great performance improvement for a use case like mine.

## Passing tests

> [!NOTE]  
> update 06/05/2024
> All tests are now passed ([see comment](https://github.com/simgrid/simgrid/pull/392#issuecomment-2093609072))

This algorithm passed all but these 4 tests:
> The following tests FAILED:
	455 - s4u-clusters-multicpu-thread (Failed)
	456 - s4u-clusters-multicpu-boost (Failed)
	457 - s4u-clusters-multicpu-raw (Failed)
	458 - s4u-clusters-multicpu-ucontext (Failed)

I've seen that there is indeed an incoherence in the route calculation, so I'm not replacing the existing algorithm. 
I do not have much time to work on it, but I think it would be interesting.

> [!NOTE]  
> update 06/05/2024
> The routes were actually valid, but when the destination was the gateway to enter in its zone, the route of the destination to itself was added (the loop did not stop). 
> Corrected by [a condition](https://github.com/JRodez/simgrid/blob/ddafba460da954dcb90898b4e91bd90ac177bca1/src/kernel/routing/NetZoneImpl.cpp#L573-L582).


mquinson:
Hello,

a merge request breaking tests have no chance of getting through, sorry about it. I understand that your time is scarce, but a merge request breaking tests is not a real *request*. 

I'm still thankful for your contribution and will try to investiguate, but this is only to say that I'm marking the PR as "draft".

Thanks, Mt

JRodez:
Hi @mquinson, [this commit](https://github.com/simgrid/simgrid/pull/392/commits/026645eae4a84470046b35eb05b30f2dba8dca60) can pass all the tests.
The code is dirty, I will find a moment for a refactoring.

JRodez:
I refactored the code and synced the repository but it seems that the new iprobe_ok check is not validated, which is verry strange because the platform used seems to have the same routes resolving between the old and new algo code.

mquinson:
Hello @JRodez, why are you including your changes in a `#ifdef` block? Your changes actually fix a bug, it should be included in all cases, or did I miss something?

We need some tests before I merge your changes. The easiest is to add the platform to use in teshsuite/platforms and then extend the flatifier.tesh file in that directory. Please try to keep the platform not too big to prevent the file from exploding in size.

And finally, don't worry if github reports that your changes don't merge, I'll take care of it.

Thanks for your work,
Mt

Issue #394, opened 2024-04-01
Bug when activity is suspended several times


Ksenia-C:
Hi, I had an asynchronous task that was suspended and resumed several times. I noticed that after calling `suspend()` for the second time, it wasn't actually suspended. If you look in the [code](https://github.com/simgrid/simgrid/blob/master/src/s4u/s4u_Activity.cpp#L153), you'll see `suspended_ = true` but in the `resume` [function](https://github.com/simgrid/simgrid/blob/master/src/s4u/s4u_Activity.cpp#L165) this variable isn't set back to `false`. As a result, at [this line](https://github.com/simgrid/simgrid/blob/master/src/s4u/s4u_Activity.cpp#L155), the function exits earlier. The fix is simple:
```
Activity* Activity::resume()
{
  if (not suspended_)
    return this; // nothing to restore when it's not suspended

  if (state_ == State::STARTED)
    pimpl_->resume();

  suspended_ = false;
  return this;
}
```
I think it is worth writing tests, but I'm not aware how they are organized in your code. It might be faster if you write them yourselves. Alternatively, you can give me pointers or docs on how to run tests, and I'll turn this issue into a merge request.

Ksenia-C:
Thank you for your answer. I've made a pr (https://github.com/simgrid/simgrid/pull/395). I'll be happy to hear feedback if I missed something or there is something more I need to do/to add.

mquinson:
Fixed by https://github.com/simgrid/simgrid/pull/395
Thanks for the report and fix!

Ksenia-C:
One last concern - how is that synchronized with framagit? Need I duplicate the merge request there?

agiersch:
Thnaks. I just synced the repository on framagit.

PR #395, opened 2024-04-08
Fix bug when suspend() after resume() didn't work

Ksenia-C:
If an activity was suspended and resumed in code several times, after the second suspend() the activity wasn't actually suspended(), and the remaining flops continued decreasing over the simulation.


Ksenia-C:
PR for the issue https://github.com/simgrid/simgrid/issues/394

mquinson:
Thanks a lot!

PR #402, opened 2024-12-25
Start providing smpi binding in python

jnquintin:


mquinson:
This is veeeeery good, thanks a lot!

